PR #12, opened 2017-09-05
Using the ClientOptions provided at App initialization to create the …

hiranya911:
Currently we use `http.DefaultClient` in the `auth` package to contact certificate server. But this [does not work](https://stackoverflow.com/questions/46038318/using-firebase-admin-sdk-for-go-from-appengine) in App Engine (and possibly some other environments). 

This PR makes it so that the client options used to initialize the App are also used to initialize the http client in `auth`. This allows users to pass in an App Engine specific http client to the App, which will then get passed into the auth module.

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -16,7 +16,7 @@ package auth
 
 import (
 	"errors"
-	"fmt"
+	"log"
 	"os"
 	"strings"
 	"testing"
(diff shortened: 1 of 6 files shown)


auth/crypto.go
```diff
@@ -29,6 +29,11 @@ import (
 	"strings"
 	"sync"
 	"time"
+
+	"golang.org/x/net/context"
```
bklimt: If you can just use "context", I think that would be better.

auth/auth_test.go
```diff
@@ -98,14 +99,17 @@ func TestMain(m *testing.M) {
 	opt := option.WithCredentialsFile("../testdata/service_account.json")
 	creds, err = transport.Creds(context.Background(), opt)
 	if err != nil {
+		fmt.Println(err)
```
bklimt: log.Fatalf() would be clearer here.

hiranya911: Done

hiranya911:
Made the suggested changes and fixed the test failure.

Turns out `transport.NewHTTPClient()` looks for application default credentials. This is not a problem in practice since most users will pass some credentials in `opts`, and others would have app default credentials present in their environment. But it does interfere with our tests when used in the Travis environment, which is not set up for app default credentials. The solution I implemented was to fall back to `http.DefaultClient` when `ctx` or `opts` are not present.

PR #18, opened 2017-09-09
Enabling the use of App Engine's SignBytes, ServiceAccount and PublicCertificates

### LINKED ISSUE #17: FR: enable custom token creation on App Engine without a JSON credentials file
In the App Engine standard environment, the code could likely be much simpler and easier to configure if it relied on [`appengine.ServiceAccount`](https://godoc.org/google.golang.org/appengine#ServiceAccount) and [`appengine.SignBytes`](https://godoc.org/google.golang.org/appengine#SignBytes) to generate custom auth tokens.

I imagine this would require a refactor to accept a `"context"` for most methods and also some additional finagling to prevent duplicate code across build tags, but it would definitely get me to lean on this library more.

An example of using the GAE tools to generate a custom token: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/docs/appengine/firebase/tictactoe/firebase.go#L45-L69

I'm definitely interested in helping out by working on a PR if there's interest in this functionality.

jprobinson:
If no additional `option.ClientOptions` are supplied, the `GOOGLE_APPLICATION_CREDENTIALS` environment variable is not set and the application is running in a Google App Engine environment, the client will now utilize the SignBytes, ServiceAccount and PublicCertificates functions provided by https://godoc.org/google.golang.org/appengine

This PR makes no breaking changes to the API and has been tested with `goapp test` on Go 1.6 and 1.8 locally and this has been tested on GAE proper.

Resolves #17


jprobinson:
I've signed it!

auth/auth_std_test.go
```diff
@@ -0,0 +1,65 @@
...
+		t.Errorf("CustomTokenWithClaims() = (%q, %v); want: (\"\", error)", token, err)
+	}
+
+	token, err = s.CustomTokenWithClaims("user1", map[string]interface{}{"foo": "bar"})
+	if token != "" || err == nil {
+		t.Errorf("CustomTokenWithClaims() = (%q, %v); want: (\"\", error)", token, err)
+	}
+}
+
+func TestNoProjectID(t *testing.T) {
```
hiranya911: I don't think this test and the one below would fail in the appengine environment. Therefore I'd rather leave then in `auth_test.go`.

jprobinson:
I've made the requested changes. Ready for review round 2.

...and sorry for the initial slop!

auth/auth_stg.go
```diff
@@ -0,0 +1,23 @@
+// +build !appengine
```
hiranya911: Possible typo in file name. Should be `auth_std.go`.

auth/signer.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
```
hiranya911: Should we just put this in `crypto.go`? I don't feel strongly about it though.

auth/auth.go
```diff
@@ -114,12 +128,10 @@ func (c *Client) CustomToken(uid string) (string, error) {
 // CustomTokenWithClaims is similar to CustomToken, but in addition to the user ID, it also encodes
 // all the key-value pairs in the provided map as claims in the resulting JWT.
 func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interface{}) (string, error) {
-	if c.email == "" {
+	iss, err := c.snr.Email()
+	if err != nil {
 		return "", errors.New("service account email not available")
```
hiranya911: Return the original error here: `return "", err`

auth/auth_test.go
```diff
@@ -102,15 +104,33 @@ func TestMain(m *testing.M) {
 		log.Fatalln(err)
 	}
 
+	ctx := context.Background()
```
hiranya911: Can we use an if-else construct here to make the code a bit more readable:

```
var (
  err error
  ctx context.Context
  ks  keySource
)

if appengine.IsDevAppServer() {
  // init ctx, ks,creds for app engine tests
} else {
 // init ctx, ks, creds for default tests
}
```

auth/auth_test.go
```diff
@@ -182,6 +202,9 @@ func TestCustomTokenInvalidCredential(t *testing.T) {
 }
 
 func TestVerifyIDToken(t *testing.T) {
+	if appengine.IsDevAppServer() {
```
hiranya911: I'm a little curious as to why this is needed. What's causing this test to fail in dev app server? If it's a legit reason, recommend adding a comment here to state that.

jprobinson:
All cleared up, ready to go!

hiranya911:
Hi @jprobinson. Can you add a comment here describing the steps for running the GAE-specific unit tests. I managed to get them to run (and pass) via `goapp test firebase.google.com/go/auth/...`. However, I also noticed that `appenging.ServiceAccount()` returns empty string in my test environment. So the minted custom tokens are actually invalid. How do I run `goapp test` with a proper service account?

My goal is to implement some integration tests around GAE support, and prevent the unit tests from having to compile against `appengine` package.

jprobinson:
Interesting. I'm not quite sure how we'd get a real email in there. 

Any test wouldn't be doing a true integration test as the local GAE server really is a thin mock of the true environment so it may not be worth the effort.

Also, if you want to try running it in a real environment. No extra options are required. (Not even urlfetch). Simply calling 'NewApp' with a GAE context and an appropriate projectID should suffice.

hiranya911:
Fair enough. Thanks @jprobinson.

Issue #25, opened 2017-09-21
API Reference link in README is a 404


ernsheong:
It currently points to https://firebase.google.com/docs/reference/admin/go/, which is a 404.

PR #26, opened 2017-09-21
Add GoDoc reference badge, fix 404 API link

ernsheong:
Fixes https://github.com/firebase/firebase-admin-go/issues/25


README.md
```diff
@@ -41,7 +42,7 @@ requests, code review feedback, and also pull requests.
 ## Documentation
 
 * [Setup Guide](https://firebase.google.com/docs/admin/setup/)
-* [API Reference](https://firebase.google.com/docs/reference/admin/go/)
+* [API Reference](http://godoc.org/github.com/firebase/firebase-admin-go)
```
hiranya911: This should point to https://godoc.org/firebase.google.com/go. Both here and above.

ernsheong:
@hiranya911 I've done it as requested. Is there a reason for this convention? Tracking purposes? Convention? Easily migrate later?

hiranya911:
@ernsheong It's just the convention. We use firebase.google.com/go as the import URL. It's easier to use the same URL everywhere else.

PR #36, opened 2017-10-30
Reusable HTTP Client API

hiranya911:
Now that we are working on implementing RTDB and user management APIs, we need some API/framework to make JSON+HTTP requests easily. Therefore I'm adding this new internal API.

Usage:
```
client := &internal.HTTPClient{HC: hc}

// Simple GET request
req := &internal.Request{
  Method: "GET",
  URL: someURL,
}
resp, err := client.Do(ctx, req)
var m map[string]interface{}
err := resp.Unmarshal(http.StatusOK, &m)
```

```
// POST request with headers
req := &internal.Request{
  Method: "POST",
  URL: someURL,
  Body: jsonMap,
  Opts: []internal.HTTPOption{internal.WithHeader("Name", "Value")},
}
resp, err := client.Do(ctx, req)
var m map[string]interface{}
err := resp.Unmarshal(http.StatusOK, &m)
```

### DIFF
--- a/internal/http_client.go
+++ b/internal/http_client.go
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package internal contains functionality that is only accessible from within the Admin SDK.
 package internal
 
 import (
--- a/internal/http_client.go
+++ b/internal/http_client.go
@@ -25,10 +25,16 @@ import (
 	"golang.org/x/net/context"
 )
 
-// HTTPClient can be used to send and receive JSON messages over HTTP.
+// HTTPClient is a convenient API to make HTTP calls.
+//
+// This API handles some of the repetitive tasks such as entity serialization and deserialization
+// involved in making HTTP calls. It provides a convenient mechanism to set headers and query
+// parameters on outgoing requests, while enforcing that an explicit context is used per request.
+// Responses returned by HTTPClient can be easily parsed as JSON, and provide a simple mechanism to
+// extract error details.
 type HTTPClient struct {
-	HC *http.Client
-	EP ErrorParser
+	Client    *http.Client
+	ErrParser ErrorParser
 }
 
 // Do executes the given Request, and returns a Response.
@@ -38,7 +44,7 @@ func (c *HTTPClient) Do(ctx context.Context, r *Request) (*Response, error) {
 		return nil, err
 	}
 
-	resp, err := c.HC.Do(req.WithContext(ctx))
+	resp, err := c.Client.Do(req.WithContext(ctx))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


hiranya911:
@avishalom please check if this API meets your requirements for implementing the user management support. I coded it with the RTDB API in mind, but I tried to keep it as general as  possible.

internal/http_client.go
```diff
@@ -0,0 +1,193 @@
...
+	Body      []byte
+	errParser ErrorParser
+}
+
+// CheckStatus checks whether the Response status code has the given HTTP status code.
+//
+// Returns an error if the status code does not match. If an ErroParser is specified, uses that to
+// construct the returned error message. Otherwise includes the full response body in the error.
+func (r *Response) CheckStatus(want int) error {
+	if r.Status == want {
```
bklimt: One concern I have about this is that there are HTTP endpoints where any 200 response is acceptable, and the client may not care at all if it's 200 or 201 or whatever. Are there cases where this CheckStatus actually needs to check some other code than a 200?

internal/http_client.go
```diff
@@ -0,0 +1,193 @@
...
+
+// Unmarshal checks if the Response has the given HTTP status code, and if so unmarshals the
+// response body into the variable pointed by v.
+//
+// Unmarshal uses https://golang.org/pkg/encoding/json/#Unmarshal internally, and hence v has the
+// same requirements as the json package.
+func (r *Response) Unmarshal(want int, v interface{}) error {
+	if err := r.CheckStatus(want); err != nil {
+		return err
+	} else if err := json.Unmarshal(r.Body, v); err != nil {
```
bklimt: I would probably leave off the redundant "else", since there's a return in there.

internal/http_client.go
hiranya911: There are cases in RTDB where we would check for 2 status codes (e.g. 200 and 412 from a conditional request), and handle the response accordingly. Something like:

```
if err := resp.CheckStatus(200); err == nil {
  // Operation successful -- Parse response
} else if err := resp.CheckStatus(412); err == nil {
  // Expected failure -- Parse error response accordingly
} else {
  // Unexpected failure
  return err
}
```

internal/http_client.go
hiranya911: Done

internal/http_client.go
```diff
@@ -0,0 +1,194 @@
...
+
+// Request contains all the parameters required to construct an outgoing HTTP request.
+type Request struct {
+	Method string
+	URL    string
+	Body   HTTPEntity
+	Opts   []HTTPOption
+}
+
+func (r *Request) newHTTPRequest() (*http.Request, error) {
```
avishalom: I have a slight preference to a name like assembleHTTPRequest or buildHTTPRequest.
newHTTPRequests hints at a constructor, but the usage involves setting up the components and calling this function on them.

internal/http_client.go
```diff
@@ -0,0 +1,194 @@
...
+	return &Response{
+		Status:    resp.StatusCode,
+		Body:      b,
+		Header:    resp.Header,
+		errParser: c.ErrParser,
+	}, nil
+}
+
+// Request contains all the parameters required to construct an outgoing HTTP request.
+type Request struct {
```
avishalom: Maybe rename RequestComponents for clarity.

hiranya911: I'd prefer to keep the name short, since the users of this API will have to type it repeatedly. Also I wanted to closely mimic the nomenclature of the http package.

internal/http_client.go
hiranya911: Done. Changed to `buildHTTPRequest()`.

PR #39, opened 2017-11-05
Fixing package path

mblasi:

### Discussion

  * Read the contribution guidelines (CONTRIBUTING.md).
  * If this has been discussed in an issue, make sure to link to the issue here. 
    If not, go file an issue about this **before creating a pull request** to discuss.

### Testing

  * Make sure all existing tests in the repository pass after your change.
  * If you fixed a bug or added a feature, add a new test to cover your code.

### API Changes

    can discuss it together.

Hi guys!

Not sure why, but it is like the context package path is wrong... either I'm loosing something or it cant work as is.

Please, let me know if the fix looks good for you, It fixes my app in my environment.

Regards,
Matías.

### DIFF
--- a/storage/storage.go
+++ b/storage/storage.go
@@ -16,8 +16,8 @@
 package storage
 
 import (
-	"context"
 	"errors"
+	"golang.org/x/net/context"
 
 	"cloud.google.com/go/storage"
 	"firebase.google.com/go/internal"


mblasi:
I signed it!

hiranya911:
`golang.org/x/net/context` and `context` refer to the same thing. The latter was introduced in golang 1.7 (see the [godocs](https://godoc.org/golang.org/x/net/context)). 

In order to support older go distributions, we use the former everywhere in our code. This is probably not necessary though, since most users are on golang 1.7 and above. But that should be done everywhere in the SDK, not just the storage package.

mblasi:
I understand, but I'm using golang 1.9.1... should it work too?

hiranya911:
It should. Our CI environment compiles and tests the SDK on go1.9 it seems: https://travis-ci.org/firebase/firebase-admin-go/builds/296402828

PR #41, opened 2017-11-21
Implemented IID Delete API

hiranya911:
A new API that enables deleting instance IDs associated with a Firebase project.

```
client, err := app.InstanceID(ctx)
if err != nil {
  // handle error
}

if err := client.DeleteInstanceID(iid); err != nil {
  // handle error
}
```

go/firebase-admin-iid

### DIFF
--- a/iid/iid.go
+++ b/iid/iid.go
@@ -29,8 +29,9 @@ import (
 
 const iidEndpoint = "https://console.firebase.google.com/v1"
 
-// Client is the interface for the Firebase instance ID service.
+// Client is the interface for the Firebase Instance ID service.
 type Client struct {
+	// To enable testing against arbitrary endpoints.
 	endpoint string
 	client   *internal.HTTPClient
 	project  string
--- a/iid/iid.go
+++ b/iid/iid.go
@@ -29,6 +29,17 @@ import (
 
 const iidEndpoint = "https://console.firebase.google.com/v1"
 
+var errorCodes = map[int]string{
+	400: "Invalid argument. Instance ID %q is malformed.",
+	401: "Request not authorized.",
+	403: "Permission denied. Project does not match instance ID or the client does not have sufficient privileges.",
+	404: "Failed to find the instance ID: %q.",
+	409: "Instance ID %q is already deleted.",
+	429: "Request throttled out by the backend server.",
+	500: "Internal server error.",
+	503: "Backend servers are over capacity. Try again later.",
+}
+
 // Client is the interface for the Firebase Instance ID service.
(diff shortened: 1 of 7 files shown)


iid/iid.go
```diff
@@ -0,0 +1,75 @@
...
+	"google.golang.org/api/transport"
+
+	"firebase.google.com/go/internal"
+
+	"golang.org/x/net/context"
+)
+
+const iidEndpoint = "https://console.firebase.google.com/v1"
+
+// Client is the interface for the Firebase instance ID service.
```
bklimt: I would capitalize "instance" here.

iid/iid.go
```diff
@@ -0,0 +1,75 @@
...
+type Client struct {
+	endpoint string
```
bklimt: Maybe add a comment that this is here only for replacing it during testing?

iid/iid_test.go
```diff
@@ -0,0 +1,139 @@
...
+	client.endpoint = ts.URL
+	if err := client.DeleteInstanceID(ctx, "test-iid"); err != nil {
+		t.Errorf("DeleteInstanceID() = %v; want nil", err)
+	}
+
+	if tr == nil {
+		t.Fatalf("Request = nil; want non-nil")
+	}
+	if tr.Method != "DELETE" {
+		t.Errorf("Method = %q; want = %q", tr.Method, "DELETE")
```
bklimt: On these, it should probably be "want %q" instead of "want = %q".

integration/iid/iid_test.go
```diff
@@ -0,0 +1,61 @@
...
+	os.Exit(m.Run())
+}
+
+func TestNonExisting(t *testing.T) {
+	err := client.DeleteInstanceID(context.Background(), "non-existing")
+	if err == nil {
+		t.Errorf("DeleteInstanceID(non-existing) = nil; want error")
+	}
+	if !strings.HasPrefix(err.Error(), "http error status: 404") {
+		t.Errorf("DeleteInstanceID(non-existing) = %v; want = 404", err)
```
bklimt: Same here.

iid/iid.go
hiranya911: Done

iid/iid.go
hiranya911: Done

iid/iid_test.go
hiranya911: We've been following this convention in this repo. I would stick to it for now. We can later discuss if we should revise it.

PR #50, opened 2017-12-08
Adding version, stat counter header and test

avishalom:
Adding a header for the go auth stat counter. 
Also a comment for cleanup.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -68,6 +68,7 @@ type Client struct {
 	projectID string
 	snr       signer
 	url       string
+	version   string
 }
 
 type signer interface {
@@ -128,15 +129,18 @@ func NewClient(ctx context.Context, c *internal.AuthConfig) (*Client, error) {
 		projectID: c.ProjectID,
 		snr:       snr,
 		url:       idToolKitURL,
+		version:   c.Version,
 	}, nil
 }
 
 // Passes the request struct, returns a byte array of the json
 func (c *Client) makeHTTPCall(ctx context.Context, serviceName string, payload interface{}, result interface{}) error {
+	statsHeader := internal.WithHeader("X-Client-Version", "Go/Admin/"+c.version)
 	request := &internal.Request{
 		Method: "POST",
 		URL:    c.url + serviceName,
 		Body:   internal.NewJSONEntity(payload),
+		Opts:   []internal.HTTPOption{statsHeader},
 	}
 	resp, err := c.hc.Do(ctx, request)
 	if err != nil {
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -129,18 +129,18 @@ func NewClient(ctx context.Context, c *internal.AuthConfig) (*Client, error) {
 		projectID: c.ProjectID,
 		snr:       snr,
 		url:       idToolKitURL,
-		version:   c.Version,
+		version:   "Go/Admin/" + c.Version,
 	}, nil
 }
 
 // Passes the request struct, returns a byte array of the json
 func (c *Client) makeHTTPCall(ctx context.Context, serviceName string, payload interface{}, result interface{}) error {
-	statsHeader := internal.WithHeader("X-Client-Version", "Go/Admin/"+c.version)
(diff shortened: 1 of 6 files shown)


auth/auth.go
```diff
@@ -128,15 +129,18 @@ func NewClient(ctx context.Context, c *internal.AuthConfig) (*Client, error) {
 		projectID: c.ProjectID,
 		snr:       snr,
 		url:       idToolKitURL,
+		version:   c.Version,
```
hiranya911: Lets assign the formatted string `Go/Admin/<version>` here, so we don't have to recompute it for each request.

auth/auth.go
```diff
@@ -128,15 +129,18 @@ func NewClient(ctx context.Context, c *internal.AuthConfig) (*Client, error) {
...
 	}, nil
 }
 
 // Passes the request struct, returns a byte array of the json
 func (c *Client) makeHTTPCall(ctx context.Context, serviceName string, payload interface{}, result interface{}) error {
+	statsHeader := internal.WithHeader("X-Client-Version", "Go/Admin/"+c.version)
```
hiranya911: versionHeader

auth/user_mgt_test.go
```diff
@@ -700,13 +725,31 @@ func echoServer(resp interface{}, t *testing.T) *mockAuthServer {
...
 	if err != nil {
-		t.Fatal()
+		t.Fatal(err)
 	}
 	authClient.url = s.Srv.URL + "/"
 	s.Client = authClient
 	return &s
 }
 
+func buildVersionTestServer(version string, t *testing.T) *mockAuthServer {
```
hiranya911: Instead of this, can we just put something permanently into the `mockAuthServer`? That way we don't have to have a new test case.

internal/http_client.go
```diff
@@ -43,7 +43,7 @@ func (c *HTTPClient) Do(ctx context.Context, r *Request) (*Response, error) {
 	if err != nil {
 		return nil, err
 	}
-
+	//
```
hiranya911: Remove

auth/user_mgt_test.go
```diff
@@ -643,6 +643,31 @@ func TestHTTPError(t *testing.T) {
 	}
 }
 
+func TestVersionHeader(t *testing.T) {
```
hiranya911: Can we try to test this without a new test case? See my comment below.

auth/user_mgt_test.go
avishalom: is this what you meant ?

auth/user_mgt_test.go
avishalom: done

auth/user_mgt_test.go
```diff
@@ -680,12 +697,14 @@ func echoServer(resp interface{}, t *testing.T) *mockAuthServer {
 
 	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		defer r.Body.Close()
-
 		reqBody, err := ioutil.ReadAll(r.Body)
 		if err != nil {
 			t.Fatal(err)
 		}
 		s.Req = append(s.Req, r)
+		if err := checkRequestsForVersionHeader(r, "Go/Admin/test.version"); err != nil {
```
hiranya911: Can this be simplified?

```
vh := r.Header.Get("X-Client-Version")
if vh != want {
   // error
}
```

auth/user_mgt_test.go
```diff
@@ -675,17 +676,20 @@ func echoServer(resp interface{}, t *testing.T) *mockAuthServer {
...
-
 		reqBody, err := ioutil.ReadAll(r.Body)
 		if err != nil {
 			t.Fatal(err)
 		}
 		s.Req = append(s.Req, r)
+		vh := r.Header.Get("X-Client-Version")
+		wantvh := "Go/Admin/" + testVersion
+		if vh != wantvh {
+			t.Errorf("version header = %s; want: %s", vh, wantvh)
```
hiranya911: Nit: Use %q instead of %s

PR #54, opened 2017-12-15
Firebase App auto init

avishalom:
Read config from the file named in FIREBASE_CONFIG env var.
ignore the existing FIREBASE_CONFIG if it is set when the tests start.


firebase.go
```diff
@@ -47,18 +49,29 @@ var firebaseScopes = []string{
 // Version of the Firebase Go Admin SDK.
 const Version = "2.2.1"
 
+// FirebaseEnvName is the name of the enviornment variable with the Config.
```
bklimt: "environment"

firebase.go
```diff
@@ -128,8 +145,43 @@ func NewApp(ctx context.Context, config *Config, opts ...option.ClientOption) (*
...
+	dat, err := ioutil.ReadFile(confFileName)
+	if err != nil {
+		return nil, err
+	}
+	err = json.Unmarshal(dat, fbc)
+	if err != nil {
+		return nil, err
+	}
+	jsonData := map[string]string{}
+	json.Unmarshal(dat, &jsonData)
```
bklimt: Please add a TODO that this code can be removed after we support Go 1.10+, by changing the previous Unmarshal to something like:

	d := json.NewDecoder(bytes.NewReader(dat))
	d.DisallowUnknownFields()
	e := d.Decode(fbc)

https://github.com/golang/go/commit/2596a0c075aeddec571cd658f748ac7a712a2b69

firebase.go
```diff
@@ -128,8 +145,43 @@ func NewApp(ctx context.Context, config *Config, opts ...option.ClientOption) (*
...
 	return &App{
 		creds:         creds,
+		databaseURL:   config.DatabaseURL,
 		projectID:     pid,
 		storageBucket: config.StorageBucket,
 		opts:          o,
 	}, nil
 }
+
+func amendDefaultConfig(config *Config) (*Config, error) {
```
bklimt: I'm not thrilled with the name "amend" here, but I don't have a great alternative. Please add a comment explaining that this fills in values missing from the passed in config (by mutating it), if they are present in the JSON file specified by the env variable.

I'm not sure it's helpful for the function to return a config, since it's always the config that was passed in. I suspect that's more likely to just be confusing. Maybe it should just return `error`?

firebase_test.go
```diff
@@ -350,6 +568,34 @@ func (t *testTokenSource) Token() (*oauth2.Token, error) {
...
+	if varSet {
+		os.Setenv(varName, oldVal)
+	} else {
+		os.Unsetenv(varName)
+	}
+}
+
+func compareConfig(got *App, want *Config, t *testing.T) {
+	if got.databaseURL != want.DatabaseURL {
+		t.Errorf("app.databaseURL = %q; want %q", got.databaseURL, want.DatabaseURL)
```
bklimt: Super nitpicky, but we've been using "want = %q" in our other tests instead of "want %q".

testdata/firebase_config.json
```diff
@@ -0,0 +1,5 @@
+{
```
bklimt: If a developer has the environment variable set to this before they run all tests, do the tests work, or will this break some of them?

testdata/firebase_config_bad_key.json
```diff
@@ -0,0 +1,3 @@
+{
+  "databaseUrl": "https://hipster-chat.firebaseio.mock"
```
bklimt: Do we really care that it fails with a case mismatch? I think json.Unmarshal usually handles case mismatches. So maybe we should actually let this case work, or at least ignore it, and test a bad key that isn't just a case difference from a good key?

firebase_test.go
```diff
@@ -350,6 +568,34 @@ func (t *testTokenSource) Token() (*oauth2.Token, error) {
 	}, nil
 }
 
+func overwriteEnv(varName, newVal string, t *testing.T) (string, bool) {
```
bklimt: To make this simpler, could we treat the empty string the same as unset? I think the amend function already does so. Then we wouldn't need the bool value.

firebase_test.go
```diff
@@ -338,6 +338,224 @@ func TestVersion(t *testing.T) {
...
+		StorageBucket: "hipster-chat.appspot.mock",
+	}
+	compareConfig(app, want, t)
+}
+
+func TestAutoInitNoEnvVar(t *testing.T) {
+	configOld, configExists := overwriteEnv(FirebaseEnvName, "", t)
+	defer reinstateEnv(FirebaseEnvName, configOld, configExists)
+
+	err := os.Unsetenv(FirebaseEnvName)
```
bklimt: If overwriteEnv treats "" as unset (as commented below), then this code can go away.

firebase_test.go
```diff
@@ -338,6 +338,224 @@ func TestVersion(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("got error = %s; wanted nil", err)
+	}
+
+	want := &Config{
+		DatabaseURL:   "https://hipster-chat.firebaseio.mock",
+		ProjectID:     "hipster-chat-mock",
+		StorageBucket: "hipster-chat.appspot.mock",
+	}
+	compareConfig(app, want, t)
```
bklimt: These tests all have the same basic structure, and I wonder if it wouldn't be clearer to do this as a table-driven test.
https://blog.golang.org/subtests

The test could be defined as a:

    struct {
        env     string
        initial &Config
        want    &Config
    }

Then the one test function could grab the env, defer reinstating it, and then then loop over each case setting the config, calling NewApp, and then comparing the result. I'm curious what @hiranya911 thinks.

testdata/firebase_config_bad_key.json
avishalom: You are right, json ignores case.  (I was thinking of catching a subtle error..)
Should I change the underlying code to ignore case as well, or just have the test case explicitly fail? 
I think I'll just change the test case, and let the TODO (disallowed) take care of that..

testdata/firebase_config.json
avishalom: We overwrite the param before each of the options tests and reinstate it after.
I've tested it locally trying both a valid and an bad filename. the tests work either way, did I miss something?

firebase_test.go
avishalom: This file has no such instances, would you like me to change the entire file (12 other instances)
(in other news I just changed wanted => want in this file.)

firebase_test.go
avishalom: Yeah, I guess if the user had explicitly set export FIREBASE_CONFIG=""
They'd like an error message.

firebase_test.go
avishalom: Yeah, I did a similar thing in the python tests, I agree it is cleaner.

firebase_test.go
avishalom: This has the downside that if for any reason the user had 
export FIREBASE_CONFIG=""
after the tests FIREBASE_CONFIG will be unset .
I am ok with this

firebase_test.go
avishalom: Iv'e change that but I think there should be a distinct test case for this

firebase_test.go
hiranya911: +1

firebase_test.go
```diff
@@ -338,6 +338,224 @@ func TestVersion(t *testing.T) {
...
+		StorageBucket: "",
+	}
+	compareConfig(app, want, t)
+}
+
+func TestAutoInitPartialOverride(t *testing.T) {
+	configOld, configExists := overwriteEnv(FirebaseEnvName, "testdata/firebase_config_partial.json", t)
+	defer reinstateEnv(FirebaseEnvName, configOld, configExists)
+
+	varName := "GOOGLE_APPLICATION_CREDENTIALS"
```
hiranya911: Define a constant for this since you're referring to it multiple times.

firebase_test.go
```diff
@@ -338,6 +338,224 @@ func TestVersion(t *testing.T) {
...
+func TestAutoInitPartialOverrideWithoutEnv(t *testing.T) {
+	configOld, configExists := overwriteEnv(FirebaseEnvName, "testdata/firebase_config_partial.json", t)
+	defer reinstateEnv(FirebaseEnvName, configOld, configExists)
+	os.Unsetenv(FirebaseEnvName)
+
+	varName := "GOOGLE_APPLICATION_CREDENTIALS"
+	credOld, credExists := overwriteEnv(varName, "testdata/service_account.json", t)
+	defer reinstateEnv(varName, credOld, credExists)
+
+	app, err := NewApp(context.Background(),
```
hiranya911: Super awkward formatting here. Init Config in a separate line.

firebase_test.go
```diff
@@ -338,6 +338,224 @@ func TestVersion(t *testing.T) {
...
+	configOld, configExists := overwriteEnv(FirebaseEnvName, "testdata/no_such_file.json", t)
+	defer reinstateEnv(FirebaseEnvName, configOld, configExists)
+
+	varName := "GOOGLE_APPLICATION_CREDENTIALS"
+	credOld, credExists := overwriteEnv(varName, "testdata/service_account.json", t)
+	defer reinstateEnv(varName, credOld, credExists)
+
+	we := "open testdata/no_such_file.json: no such file or directory"
+
+	_, err := NewApp(context.Background(), &Config{DatabaseURL: "d", ProjectID: "p", StorageBucket: "s"})
```
hiranya911: Is this the behavior we want? I thought we implemented different semantics in Node?

firebase.go
```diff
@@ -128,8 +145,43 @@ func NewApp(ctx context.Context, config *Config, opts ...option.ClientOption) (*
...
+		config.DatabaseURL = fbc.DatabaseURL
+	}
+	if config.ProjectID == "" {
+		config.ProjectID = fbc.ProjectID
+	}
+	if config.StorageBucket == "" {
+		config.StorageBucket = fbc.StorageBucket
+	}
+
+	return config, nil
```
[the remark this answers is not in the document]
hiranya911: Suggested algorithm (in pseudo code):

```
devConf = conf.toMap()
envConf = fbc.toMap()
for k, _ := range envConf {
  // check k is allowed
}
for k, _ := range validFieldNames {
  v := devConf[k]
  if notZero(v) {
     setField(fbc, k, v) // Probably using reflection?
  }
}
return fbc
```

Thoughts?

firebase_test.go
avishalom: deleted in a previous commit.

testdata/firebase_config.json
bryanklimt: What I meant was that setting this env var will affect the options used by other tests (for other go admin packages) that might depend on the assumption that the config defaults to empty. But it's probably fine.

bklimt: What I meant was that setting this env var will affect the options used by other tests (for other go admin packages) that might depend on the assumption that the config defaults to empty. But it's probably fine.

firebase.go
```diff
@@ -133,3 +147,41 @@ func NewApp(ctx context.Context, config *Config, opts ...option.ClientOption) (*
 		opts:          o,
 	}, nil
 }
+
+// amendConfigWithDefaults reads the default config file, defined by the FIREBASE_COFIG
```
bklimt: s/COFIG/CONFIG

internal/internal.go
```diff
@@ -50,3 +53,25 @@ type MockTokenSource struct {
 func (ts *MockTokenSource) Token() (*oauth2.Token, error) {
 	return &oauth2.Token{AccessToken: ts.AccessToken}, nil
 }
+
+// OverwriteEnv ovevrwrites env variables, used in testsing.
```
bklimt: s/ovevrwrites/overwrites

testdata/firebase_config.json
avishalom: You are right, my test was wrong, if I add the variable locally, it fails those tests that that exist and do not set it. If guarded against this by wrapping the set and restore in the TestMain of each package, I will add this to the style guide. (i.e. when checking the integration tests, you should run them with and without the setting of the env vars)

firebase.go
```diff
@@ -128,8 +145,43 @@ func NewApp(ctx context.Context, config *Config, opts ...option.ClientOption) (*
...
+		return nil, err
+	}
+	jsonData := map[string]string{}
+	json.Unmarshal(dat, &jsonData)
+	for k := range jsonData {
+		if _, ok := validConfigFieldNames[k]; !ok {
+			return nil, fmt.Errorf(`unexpected field %s in JSON config file`, k)
+		}
+	}
+	if config.DatabaseURL == "" {
```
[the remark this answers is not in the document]
avishalom: Done, not to verbose.

firebase.go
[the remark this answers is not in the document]
avishalom: Done with reflect

avishalom:
Done. Will leave the want = changes for the next PR as discuss offline.
Let me know what you think of the reflection copy, at this stage we can omit the int switch case.

firebase.go
```diff
@@ -107,14 +111,16 @@ func (a *App) InstanceID(ctx context.Context) (*iid.Client, error) {
...
 	o = append(o, opts...)
-
 	creds, err := transport.Creds(ctx, o...)
 	if err != nil {
 		return nil, err
 	}
-
 	if config == nil {
-		config = &Config{}
+		config, err = getConfigDefaults()
```
bklimt: I would put these lines together:

    if config, err = getConfigDefaults(); err != nil {
        return nil, err
    }

firebase.go
```diff
@@ -133,3 +139,25 @@ func NewApp(ctx context.Context, config *Config, opts ...option.ClientOption) (*
...
+	confFileName := os.Getenv(firebaseEnvName)
+	if confFileName == "" {
+		return fbc, nil
+	}
+	var dat []byte
+	if confFileName[0] == byte('{') {
+		dat = []byte(confFileName)
+	} else {
+		var err error
+		dat, err = ioutil.ReadFile(confFileName)
```
bklimt: I'd combine these two lines the same way.

firebase_test.go
```diff
@@ -337,6 +347,165 @@ func TestVersion(t *testing.T) {
...
 }
+func TestAutoInit(t *testing.T) {
+	tests := []struct {
+		name         string
+		confFilename string
+		initOptions  *Config
+		wantOptions  *Config
+	}{
+		{
+			"no environment var, no options",
```
hiranya911: Please revise the names of these test cases. Some of them are outdated/invalid considering the changes we've done. Also lets use meaningful sentences like:

* No environment variable, and no explicit options
* Environment variable set to file, and no explicit options
* Environment variable set to string, and no explicit options

etc

PR #55, opened 2017-12-20
Removing redundant if

avishalom:
removing redundant check to placate the linter.


hiranya911:
Note that this is currently not against the dev branch. Be mindful when merging.

PR #57, opened 2017-12-21
Release Version 2.3.0

hiranya911:


auth/auth.go
```diff
@@ -102,6 +101,7 @@ func NewClient(ctx context.Context, c *internal.AuthConfig) (*Client, error) {
 		}
 		email = svcAcct.ClientEmail
 	}
+
```
avishalom: may as well keep the spacing

auth/auth.go
```diff
@@ -111,21 +111,15 @@ func NewClient(ctx context.Context, c *internal.AuthConfig) (*Client, error) {
...
 	}
-	hc := http.DefaultClient
-	if len(c.Opts) > 0 { // TODO: fix the default when len = 0
-		hc, _, err = transport.NewHTTPClient(ctx, c.Opts...)
-		if err != nil {
-			return nil, err
-		}
-	}
-	ks, err := newHTTPKeySource(googleCertURL, hc)
+
```
avishalom: ditto

auth/auth.go
hiranya911: This is just a simple merge of the existing dev branch into master. We don't make manual changes here. These have to be addressed on dev as separate PRs.

PR #58, opened 2017-12-26
use const for Common HTTP methods and HTTP status codes

chemidy:
Just small changes to use const for Common HTTP methods and HTTP status codes instead of "200" and "GET"

### DIFF
--- a/firebase_test.go
+++ b/firebase_test.go
@@ -85,8 +85,8 @@ func TestClientOptions(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if resp.StatusCode != 200 {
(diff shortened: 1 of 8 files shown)


firebase_test.go
```diff
@@ -85,8 +85,8 @@ func TestClientOptions(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if resp.StatusCode != 200 {
-		t.Errorf("Status: %d; want: 200", resp.StatusCode)
+	if resp.StatusCode != http.StatusOK {
+		t.Errorf("Status: %d; want: %q", resp.StatusCode, http.StatusOK)
```
hiranya911: `want: %d`

PR #60, opened 2017-12-29
use identitytoolkit/v3 go client instead of makeHTTPCall

chemidy:
Replace  of makeHTTPCall by [google-api-go-client identitytoolkit/v3](https://github.com/google/google-api-go-client/tree/master/identitytoolkit/v3)

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -22,17 +22,16 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
-	"net/http"
 	"strings"
 
 	"firebase.google.com/go/internal"
 	"golang.org/x/net/context"
+	"google.golang.org/api/identitytoolkit/v3"
 	"google.golang.org/api/transport"
 )
 
 const firebaseAudience = "https://identitytoolkit.googleapis.com/google.identity.identitytoolkit.v1.IdentityToolkit"
 const googleCertURL = "https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com"
-const idToolKitURL = "https://www.googleapis.com/identitytoolkit/v3/relyingparty/"
 const issuerPrefix = "https://securetoken.google.com/"
 const tokenExpSeconds = 3600
 
@@ -64,6 +63,7 @@ type Token struct {
 // by Firebase backend services.
 type Client struct {
 	hc        *internal.HTTPClient
+	is        *identitytoolkit.Service
 	ks        keySource
 	projectID string
 	snr       signer
@@ -118,30 +118,19 @@ func NewClient(ctx context.Context, c *internal.AuthConfig) (*Client, error) {
 		return nil, err
 	}
 
+	is, err := identitytoolkit.New(hc)
+	if err != nil {
+		return nil, err
+	}
+
 	return &Client{
 		hc:        &internal.HTTPClient{Client: hc},
+		is:        is,
 		ks:        newHTTPKeySource(googleCertURL, hc),
 		projectID: c.ProjectID,
 		snr:       snr,
-		url:       idToolKitURL,
 		version:   "Go/Admin/" + c.Version,
 	}, nil
-}
-
-// Passes the request struct, returns a byte array of the json
-func (c *Client) makeHTTPCall(ctx context.Context, serviceName string, payload interface{}, result interface{}) error {
-	versionHeader := internal.WithHeader("X-Client-Version", c.version)
-	request := &internal.Request{
-		Method: http.MethodPost,
-		URL:    c.url + serviceName,
-		Body:   internal.NewJSONEntity(payload),
-		Opts:   []internal.HTTPOption{versionHeader},
-	}
-	resp, err := c.hc.Do(ctx, request)
-	if err != nil {
-		return err
-	}
-	return resp.Unmarshal(http.StatusOK, result)
 }
 
 // CustomToken creates a signed custom authentication token with the specified user ID. The resulting
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -67,7 +67,6 @@ type Client struct {
 	ks        keySource
 	projectID string
 	snr       signer
-	url       string
 	version   string
 }
 
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -40,12 +40,12 @@ var commonValidators = map[string]func(interface{}) error{
 }
 
 // Create a new interface
-type headerable interface {
+type relayingpartyCall interface {
 	Header() http.Header
 }
 
 // set header
-func (c *Client) execute(ic headerable) {
+func (c *Client) setHeader(ic relayingpartyCall) {
 	ic.Header().Set("X-Client-Version", c.version)
 }
 
@@ -199,7 +199,7 @@ func (c *Client) DeleteUser(ctx context.Context, uid string) error {
 	}
 
 	relyingpartyDeleteAccountCall := c.is.Relyingparty.DeleteAccount(request)
-	c.execute(relyingpartyDeleteAccountCall)
+	c.setHeader(relyingpartyDeleteAccountCall)
 	_, err := relyingpartyDeleteAccountCall.Context(ctx).Do()
 	return err
 }
@@ -261,7 +261,7 @@ func (it *UserIterator) fetch(pageSize int, pageToken string) (string, error) {
 		NextPageToken: pageToken,
 	}
 	relyingpartyDownloadAccountCall := it.client.is.Relyingparty.DownloadAccount(request)
-	it.client.execute(relyingpartyDownloadAccountCall)
+	it.client.setHeader(relyingpartyDownloadAccountCall)
 	resp, err := relyingpartyDownloadAccountCall.Context(it.ctx).Do()
 	if err != nil {
 		return "", err
@@ -540,7 +540,7 @@ func (c *Client) createUser(ctx context.Context, user *UserToCreate) (string, er
 	}
 
 	relyingpartySignupNewUserCall := c.is.Relyingparty.SignupNewUser(request)
-	c.execute(relyingpartySignupNewUserCall)
+	c.setHeader(relyingpartySignupNewUserCall)
 	resp, err := relyingpartySignupNewUserCall.Context(ctx).Do()
 	if err != nil {
 		return "", err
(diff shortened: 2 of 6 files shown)


chemidy:
I signed it!

auth/user_mgt.go
```diff
@@ -400,12 +416,26 @@ func (u *UserToCreate) preparePayload() (map[string]interface{}, error) {
...
+		if !user.EmailVerified {
+			user.ForceSendFields = append(user.ForceSendFields, "EmailVerified")
+		}
+	}
+
 	return params, nil
 }
 
-func (u *UserToUpdate) preparePayload() (map[string]interface{}, error) {
+func (u *UserToUpdate) preparePayload(user *identitytoolkit.IdentitytoolkitRelyingpartySetAccountInfoRequest) (map[string]interface{}, error) {
```
avishalom: The first is always ignored, might as well omit it here. (and in the parallel function)

auth/user_mgt.go
```diff
@@ -468,15 +522,21 @@ func (c *Client) createUser(ctx context.Context, user *UserToCreate) (string, er
...
+
+	_, err := user.preparePayload(request)
 	if err != nil {
 		return "", err
 	}
-	var rur responseUserRecord
-	if err := c.makeHTTPCall(ctx, "signupNewUser", payload, &rur); err != nil {
+
+	relyingpartySignupNewUserCall := c.is.Relyingparty.SignupNewUser(request)
+	relyingpartySignupNewUserCall.Header().Set("X-Client-Version", c.version)
```
avishalom: I don't like the fact that the header is set in 5 different places. 
(I.e. if we ever change the header, or want to add new identifiers to every call, we'd like to do it in one place)
I don't see a super elegant way to fix this though.

maybe something along the lines of:
(this isn't ideal, ideally I'd also refactor in the next line (and add "Do() (interface{}, error)", and "Context(ctx) *\<interfaceName\>" to the new interface), but returning an interface{} from Do() and casting it later seems ugly.)
```
// Create a new interface
type headerable interface {
    Header() http.Header
}

// refactor this lines 
func (c *Client)execute(ic *headerable) (){
    ic.Header().Set("X-Client-Version", c.version)
}
```
WDYT? do you have another solution?

auth/user_mgt.go
chemidy: Done

auth/user_mgt.go
chemidy: I think it's a good compromise, i was looking for a better solution, but i didn't find anything more elegant :(

auth/user_mgt.go
```diff
@@ -38,6 +39,16 @@ var commonValidators = map[string]func(interface{}) error{
 	"localId":     validateUID,
 }
 
+// Create a new interface
+type headerable interface {
```
avishalom: Thanks for this change!
and I apologize for this secondary nit pick, but I don't like the example names I suggested.
execute would have been a good name if we rolled the ".Do()" line into it. 
as it is maybe "prepareCall", or "setHeader", or "beforeCall" or a similar name that describes the fact that this is a needed step before the call, that adds some params to the header... 

Likewise maybe instead of headerable "relayingpartyCall" would be a better descriptor. 
(again my apologies for not thinking of these names when I suggested the change.)

avishalom:
LGTM with one last small nit about names, Pending Hiranya's LGTM

auth/user_mgt.go
chemidy: OK, my first idea was setHeader so i keep it.

auth/user_mgt.go
```diff
@@ -37,6 +39,16 @@ var commonValidators = map[string]func(interface{}) error{
 	"localId":     validateUID,
 }
 
+// Create a new interface
+type relayingpartyCall interface {
```
hiranya911: Recommend calling this `identitytoolkitCall`

auth/user_mgt.go
```diff
@@ -230,25 +256,26 @@ func (c *Client) Users(ctx context.Context, nextPageToken string) *UserIterator
...
-	if pageToken != "" {
-		params["nextPageToken"] = pageToken
+	request := &identitytoolkit.IdentitytoolkitRelyingpartyDownloadAccountRequest{
+		MaxResults:    int64(pageSize),
+		NextPageToken: pageToken,
 	}
-
-	var lur listUsersResponse
-	err := it.client.makeHTTPCall(it.ctx, "downloadAccount", params, &lur)
+	relyingpartyDownloadAccountCall := it.client.is.Relyingparty.DownloadAccount(request)
```
hiranya911: Rename to `call`? Go advocates shorter names for local variables.

auth/user_mgt.go
```diff
@@ -415,17 +456,41 @@ func (u *UserToUpdate) preparePayload() (map[string]interface{}, error) {
...
+		user.CustomAttributes = params["customAttributes"].(string)
 	}
 
 	for key, validate := range commonValidators {
 		if v, ok := params[key]; ok {
 			if err := validate(v); err != nil {
-				return nil, err
+				return err
 			}
+			reflect.ValueOf(user).Elem().FieldByName(strings.Title(key)).SetString(params[key].(string))
```
hiranya911: How will this work if we had some non-string params?

auth/user_mgt.go
```diff
@@ -525,27 +602,45 @@ func makeExportedUser(r responseUserRecord) (*ExportedUserRecord, error) {
 		}
 	}
 
+	var providerID string
+	if len(r.ProviderUserInfo) == 1 {
+		providerID = r.ProviderUserInfo[0].ProviderId
```
hiranya911: @avishalom can you verify this logic is correct? Looks a little strange.

auth/user_mgt_test.go
```diff
@@ -49,9 +50,11 @@ var testUser = &UserRecord{
 			DisplayName: "Test User",
 			PhotoURL:    "http://www.example.com/testuser/photo.png",
 			Email:       "testuser@example.com",
+			UID:         "testuser@example.com",
```
hiranya911: What is the intention here? Setting UID to email (and phone number below) looks unusual.

auth/user_mgt_test.go
```diff
@@ -683,6 +705,10 @@ func echoServer(resp interface{}, t *testing.T) *mockAuthServer {
 			t.Fatal(err)
 		}
 		s.Rbody = reqBody
+		// remove extra newline for  reflect.DeepEqual
+		if string(s.Rbody[len(s.Rbody)-1]) == "\n" {
```
hiranya911: `strings.TrimSpace(s.Rbody)`

auth/user_mgt_test.go
```diff
@@ -576,32 +595,35 @@ func TestInvalidDeleteUser(t *testing.T) {
...
+		CustomAttributes: `{"admin": true, "package": "gold"}`,
+		ProviderUserInfo: []*identitytoolkit.UserInfoProviderUserInfo{
 			{
-				ProviderID:  "password",
+				ProviderId:  "password",
 				DisplayName: "Test User",
-				PhotoURL:    "http://www.example.com/testuser/photo.png",
+				PhotoUrl:    "http://www.example.com/testuser/photo.png",
 				Email:       "testuser@example.com",
+				RawId:       "testuser@example.com",
```
hiranya911: Ah I see why you added the LocalId property above. I'd recommend changing this to `testuid` so the intention is clearer.

auth/auth.go
```diff
@@ -118,32 +118,21 @@ func NewClient(ctx context.Context, c *internal.AuthConfig) (*Client, error) {
...
+		return nil, err
+	}
+
 	return &Client{
 		hc:        &internal.HTTPClient{Client: hc},
+		is:        is,
 		ks:        newHTTPKeySource(googleCertURL, hc),
 		projectID: c.ProjectID,
 		snr:       snr,
-		url:       idToolKitURL,
```
hiranya911: Can we also remove url from the definition of `Client` type?

chemidy: Yes of course

auth/user_mgt.go
chemidy: not easy to find the right name :)

auth/user_mgt.go
chemidy: perfect 👍

auth/user_mgt_test.go
chemidy: Thks

auth/user_mgt.go
chemidy: It's a mistake, i will fix it. 
This is Because on the UserRecord, the UserInfo struct is used for ProviderUserInfo and for  UserInfo. But providerID is only set for a ProviderUserInfo.

````
// UserRecord contains metadata associated with a Firebase user account.
type UserRecord struct {
	*UserInfo
	CustomClaims     map[string]interface{}
	Disabled         bool
	EmailVerified    bool
	ProviderUserInfo []*UserInfo
	UserMetadata     *UserMetadata
}
```

auth/user_mgt.go
```diff
@@ -525,27 +602,39 @@ func makeExportedUser(r responseUserRecord) (*ExportedUserRecord, error) {
...
+	}
+
 	resp := &ExportedUserRecord{
 		UserRecord: &UserRecord{
 			UserInfo: &UserInfo{
 				DisplayName: r.DisplayName,
 				Email:       r.Email,
 				PhoneNumber: r.PhoneNumber,
-				PhotoURL:    r.PhotoURL,
-				ProviderID:  r.ProviderID,
```
hiranya911: We should actually set ProviderID to the constant string `firebase` here. That's what our other Admin SDKs do, and it's a documented behavior. @avishalom can you make the necessary changes in a follow up PR and update the tests?

auth/user_mgt_test.go
chemidy: ok fix

auth/user_mgt.go
chemidy: For now in commonValidators there are only string params.
And all validators in the list cast value to string

```
func validateDisplayName(val interface{}) error {
	if len(val.(string)) == 0 {
```

This is the same behavior than if we pass a boolean to password field without using the setter.

```
&UserToCreate{params: map[string]interface{}{"password": true}},
```

instead of

```
(&UserToCreate{}).Password("123456"),
```

**panic: interface conversion: interface {} is bool, not string**

@hiranya911 Any Ideas ?

hiranya911: I was mostly concerned about any non-string attributes we may introduce in the future. But I think for now what you have done is ok. I'll leave it to @avishalom to make the final call.

avishalom: I'm happy to leave it like that and to add a type switch in the future if necessary.

auth/user_mgt.go
avishalom: Does it make sense to add a test that would have failed in this case?

avishalom:
Resolve conflicts and it looks good to me. 
(if it makes sense to add a test that would have failed on the ProviderUserInfo issue, then add a test for that.)

PR #77, opened 2018-01-25
Revoke Tokens

avishalom:
go implementation of token revocation


auth/auth.go
```diff
@@ -177,6 +177,17 @@ func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interfac
 	return encodeToken(c.snr, defaultHeader(), payload)
 }
 
+// RevokeRefreshToken revokes all refresh tokens for the specified user identified by the uid provided.
+// In addition to revoking all refresh tokens for a user, all ID tokens issued
+// before revocation will also be revoked on the Auth backend. Any request with an
+// ID token generated before revocation will be rejected with a token expired error.
+// Note that due to the fact that the timestamp is stored in seconds, any tokens minted in
+// the same second as the revocation will still be valid. If there is a chance that a token
+//  was minted in the last second, delay for 1 second before revoking.all tokens minted before the current second.
```
bklimt: There's a weird space at the beginning of this line.

auth/auth.go
```diff
@@ -177,6 +177,17 @@ func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interfac
 	return encodeToken(c.snr, defaultHeader(), payload)
 }
 
+// RevokeRefreshToken revokes all refresh tokens for the specified user identified by the uid provided.
+// In addition to revoking all refresh tokens for a user, all ID tokens issued
+// before revocation will also be revoked on the Auth backend. Any request with an
+// ID token generated before revocation will be rejected with a token expired error.
+// Note that due to the fact that the timestamp is stored in seconds, any tokens minted in
```
bklimt: Is the timestamp second granularity part of this client SDK, or is it a limitation of the backend API being called?

auth/auth.go
```diff
@@ -237,6 +248,25 @@ func (c *Client) VerifyIDToken(idToken string) (*Token, error) {
 	return p, nil
 }
 
+// VerifyIDTokenWithCheckRevoked verifies the signature	and payload of the provided ID token and
```
bklimt: weird comment spacing

auth/auth.go
```diff
@@ -237,6 +248,25 @@ func (c *Client) VerifyIDToken(idToken string) (*Token, error) {
...
+// if requested, whether it was revoked.
+// see: VerifyIDToken above.
+func (c *Client) VerifyIDTokenWithCheckRevoked(ctx context.Context, idToken string, checkToken bool) (*Token, error) {
```
bklimt: It seems odd to have both of these pieces of functionality in one function. `VerifyIdToken` is already exposed as an API people could use any time they would otherwise call this with `checkToken = false`. So why not remove the `checkToken` parameter and always assume it's true? Then maybe the function could be named something like `VerifyIDTokenNotRevoked`. TBH, I don't really understand why `VerifyIDToken` doesn't already check the revoked state, though.

auth/user_mgt.go
```diff
@@ -416,6 +424,13 @@ func validatePhone(val interface{}) error {
 	return nil
 }
 
+func validateValidSince(val interface{}) error {
+	if _, ok := val.(int64); !ok {
+		return fmt.Errorf("tokens valid after time must be an integer")
```
bklimt: I would either write this as:
"TokensValidAfterTime must be an integer"
or
"valid-after time for tokens must be an integer"

As it is, it's hard to read, because it's not clear that "tokens valid after time" is a compound, especially since the field is called "validSince" in other parts of the API.

Actually, why is it "validSince" in some places and "TokensValidAfterTime" in another?

integration/auth/user_mgt_test.go
```diff
@@ -65,6 +66,14 @@ func testCreateUsers(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	// make sure that the user.TokensValidAvterTime is not in the future or stale.
```
bklimt: s/Avter/After/

auth/auth.go
avishalom: The backend stores it as seconds since the epoch.

auth/user_mgt.go
avishalom: This is part of what was decided in the node api.

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 # Unreleased
 - 
 
+### Token revokaction
+- [added] The ['VerifyIDToken(...)'](https://godoc.org/firebase.google.com/go/auth#Client.VerifyIDToken)
+  method can now take an additional parameter `check_revoked`, when `True`,
```
hiranya911: Not accurate for Go. Please revise.

auth/auth.go
```diff
@@ -177,6 +177,17 @@ func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interfac
 	return encodeToken(c.snr, defaultHeader(), payload)
 }
 
+// RevokeRefreshToken revokes all refresh tokens for the specified user identified by the uid provided.
+// In addition to revoking all refresh tokens for a user, all ID tokens issued
+// before revocation will also be revoked on the Auth backend. Any request with an
```
hiranya911: at the Auth backend.

auth/auth.go
hiranya911: Low level implementation detail. No need to mention it here IMO.

auth/auth.go
```diff
@@ -237,6 +248,25 @@ func (c *Client) VerifyIDToken(idToken string) (*Token, error) {
...
+
+	if !checkToken || err != nil {
+		return p, err
+	}
+	user, err := c.GetUser(ctx, p.UID)
+	if err != nil {
+		return p, err
+	}
+	if p.IssuedAt < user.TokensValidAfterTime {
+		err = fmt.Errorf("the Firebase ID token has been revoked")
```
hiranya911: `id token has been revoked` (all lower case)?

hiranya911: Should also `p = nil` here.

auth/user_mgt.go
```diff
@@ -173,6 +176,11 @@ func (u *UserToUpdate) PhoneNumber(phone string) *UserToUpdate { u.set("phoneNum
 // PhotoURL setter.
 func (u *UserToUpdate) PhotoURL(url string) *UserToUpdate { u.set("photoUrl", url); return u }
 
+func (u *UserToUpdate) revokeRefreshToken() *UserToUpdate {
```
hiranya911: Should this be exposed as `TokensValidAfterTime(since int64)`?

auth/user_mgt.go
```diff
@@ -416,6 +424,13 @@ func validatePhone(val interface{}) error {
 	return nil
 }
 
+func validateValidSince(val interface{}) error {
+	if _, ok := val.(int64); !ok {
+		return fmt.Errorf("tokens valid after time must be an integer")
+	}
```
hiranya911: Also check for non-negative.

auth/auth_test.go
```diff
@@ -193,6 +192,60 @@ func TestCustomTokenInvalidCredential(t *testing.T) {
 	}
 }
 
+func TestVerifyIDTokenWithCheckRevokedValid(t *testing.T) {
+	s := echoServer(testGetUserResponse, t)
+	defer s.Close()
+
+	ft, err := s.Client.VerifyIDTokenWithCheckRevoked(ctx, testIDToken, true)
```
hiranya911: Add a test case where validSince is in the past.

integration/auth/auth_test.go
```diff
@@ -77,6 +78,57 @@ func TestCustomToken(t *testing.T) {
 	}
 }
 
+func TestCustomTokenVerifyCheckRevoked(t *testing.T) {
+	revokedID := "user_revoked"
```
hiranya911: uid

integration/auth/auth_test.go
```diff
@@ -77,6 +78,57 @@ func TestCustomToken(t *testing.T) {
...
+	_, err = client.VerifyIDTokenWithCheckRevoked(ctx, idt, false)
+	if err != nil {
+		t.Errorf("VerifyIDTokenWithCheckRevoked(.., false); err = %s; want err = <nil>", err)
+	}
+	err = client.DeleteUser(ctx, revokedID)
+	if err != nil {
+		t.Error(err)
+	}
+
+	// Sign in after revocation after revocation.
```
hiranya911: Messed up comment?

integration/auth/user_mgt_test.go
```diff
@@ -65,6 +66,14 @@ func testCreateUsers(t *testing.T) {
...
+	if u.TokensValidAfterTime > time.Now().Unix() {
+		t.Errorf("timestamp cannot be in the future")
+	}
+	if u.TokensValidAfterTime < time.Now().Unix()-3600 {
```
hiranya911: This is a new user right? What is the default TokensValidAfterTime for a new user?

auth/auth.go
avishalom: validSince is initialized to 0 when creating a new user, and we are only introducing revoke here. so we probably will not break anything if we assume it to be true.

integration/auth/user_mgt_test.go
avishalom: 0

auth/auth.go
avishalom: I feel that it should be mentioned somewhere. it would be a hard to find error if it just happened. especially as we are returning milliseconds.

auth/user_mgt.go
avishalom: I've opened it up for discussion (whether we want the devs to do something like revoke 5 minutes from now, or revoke the old ones before last Tuesday.)

auth/auth_test.go
avishalom: This test has validSince before the token 'iat'
TestVerifyIDTokenWithCheckRevokedInvalidated puts the token 'iat' before the valid since, is that what you meant?

auth/user_mgt.go
avishalom: if 1971 is an ok date, why not 1969 ?
(should we add a check that the epoch seconds are e.g. > 1e9 (~Y2K))

auth/auth.go
```diff
@@ -177,6 +177,16 @@ func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interfac
 	return encodeToken(c.snr, defaultHeader(), payload)
 }
 
+// RevokeRefreshTokens revokes all refresh tokens for the specified user.
+// In addition to revoking all refresh tokens for a user, all ID tokens issued
+// before revocation will also be revoked at the Auth backend. Any request with an
+// ID token generated before revocation will be rejected with a token expired error.
+// Note that any tokens minted in the same epoch second as the revocation will still be valid.
+// If there is a chance that a token was minted in the last second, delay for 1 second before revoking.
+func (c *Client) RevokeRefreshTokens(ctx context.Context, uid string) error {
```
hiranya911: Comment format for Go functions is something like:

```
// Single-lined overview of the function.
//
// Additional details as separate paragraphs.
```

Please format accordingly.

 I still think the race condition is an implementation detail that is not worth documenting. In fact it is going to confuse a lot of developers. In reality the chances of somebody running into this outside tests is almost non-existant.

auth/auth.go
```diff
@@ -237,6 +247,27 @@ func (c *Client) VerifyIDToken(idToken string) (*Token, error) {
 	return p, nil
 }
 
+// VerifyIDTokenWithCheckRevoked verifies the signature and payload of the provided ID token and
+// checks that it wasn't revoked.
+// see: VerifyIDToken above.
```
hiranya911: Uses VerifyIDToken() internally to verify the ID token JWT.

auth/auth.go
```diff
@@ -237,6 +247,27 @@ func (c *Client) VerifyIDToken(idToken string) (*Token, error) {
...
+func (c *Client) VerifyIDTokenWithCheckRevoked(ctx context.Context, idToken string) (*Token, error) {
+	p, err := c.VerifyIDToken(idToken)
+
```
hiranya911: Remove

auth/auth_test.go
```diff
@@ -193,6 +192,65 @@ func TestCustomTokenInvalidCredential(t *testing.T) {
...
+func TestVerifyIDTokenWithCheckRevokedDoNotCheck(t *testing.T) {
+	s := echoServer(testGetUserResponse, t)
+	defer s.Close()
+	u, err := s.Client.GetUser(ctx, "testuser")
+	if err != nil {
+		t.Fatal("Error retrieving user")
+	}
+
+	ft, err := s.Client.VerifyIDToken(
+		getIDToken(mockIDTokenPayload{
```
hiranya911: Call getIDToken on a separate line for clarity.

auth/auth_test.go
```diff
@@ -193,6 +192,65 @@ func TestCustomTokenInvalidCredential(t *testing.T) {
...
+func TestVerifyIDTokenWithCheckRevokedInvalidated(t *testing.T) {
+	s := echoServer(testGetUserResponse, t)
+	defer s.Close()
+	u, err := s.Client.GetUser(ctx, "testuser")
+	if err != nil {
+		t.Fatal("Error retrieving user")
+	}
+	p, err := s.Client.VerifyIDTokenWithCheckRevoked(
+		ctx,
+		getIDToken(mockIDTokenPayload{
```
hiranya911: Same here

auth/auth_test.go
```diff
@@ -193,6 +192,65 @@ func TestCustomTokenInvalidCredential(t *testing.T) {
...
+	}
+	if ft.UID != ft.Subject {
+		t.Errorf("UID = %q; Sub = %q; want UID = Sub", ft.UID, ft.Subject)
+	}
+}
+
+func TestVerifyIDTokenWithCheckRevokedDoNotCheck(t *testing.T) {
+	s := echoServer(testGetUserResponse, t)
+	defer s.Close()
+	u, err := s.Client.GetUser(ctx, "testuser")
```
hiranya911: This part is unnecessary IMO

auth/auth_test.go
```diff
@@ -193,6 +192,65 @@ func TestCustomTokenInvalidCredential(t *testing.T) {
...
+	}
+	p, err := s.Client.VerifyIDTokenWithCheckRevoked(
+		ctx,
+		getIDToken(mockIDTokenPayload{
+			"uid": "uid",
+			"iat": int(u.TokensValidAfterTime/1000) - 10,
+		}))
+	we := "id token has been revoked"
+	if p != nil || err == nil || err.Error() != we {
+		t.Errorf("VerifyIDTokenWithCheckRevoked(..., token) =(%v, %v); want = (%v, %v)",
```
hiranya911: Remove `...`

auth/auth_test.go
```diff
@@ -193,6 +192,65 @@ func TestCustomTokenInvalidCredential(t *testing.T) {
...
+	}
+	if ft.UID != ft.Subject {
+		t.Errorf("UID = %q; Sub = %q; want UID = Sub", ft.UID, ft.Subject)
+	}
+}
+
+func TestVerifyIDTokenWithCheckRevokedInvalidated(t *testing.T) {
+	s := echoServer(testGetUserResponse, t)
+	defer s.Close()
+	u, err := s.Client.GetUser(ctx, "testuser")
```
hiranya911: Remove

auth/user_mgt.go
```diff
@@ -73,11 +76,12 @@ type UserMetadata struct {
...
-	CustomClaims     map[string]interface{}
-	Disabled         bool
-	EmailVerified    bool
-	ProviderUserInfo []*UserInfo
-	UserMetadata     *UserMetadata
+	CustomClaims         map[string]interface{}
+	Disabled             bool
+	EmailVerified        bool
+	ProviderUserInfo     []*UserInfo
+	TokensValidAfterTime int64 // Epoch milliseconds
```
hiranya911: milliseconds since epoch

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Unreleased
 - 
 
+### Token revokaction
+- [added] A New ['VerifyIDTokenWithCheckRevoked(ctx, token)'](https://godoc.org/firebase.google.com/go/auth#Client.VerifyIDToken)
+  method checks to see if the token has been revoked. A new method ['RevokeRefreshTokens(uid)'](https://godoc.org/firebase.google.com/go/auth#Client.RevokeRefreshTokens)
```
hiranya911: Break these into two bullet points.

auth/user_mgt.go
```diff
@@ -416,6 +427,14 @@ func validatePhone(val interface{}) error {
 	return nil
 }
 
+func validateValidSince(val interface{}) error {
+	if v, ok := val.(int64); !ok || v > 1e12 {
```
hiranya911: Is this to validate the UpdateUser() call? If so it should accept a milliseconds timestamp and convert it into seconds.

auth/user_mgt_test.go
```diff
@@ -496,6 +498,43 @@ func TestUpdateUser(t *testing.T) {
...
+func TestRevokeRefreshTokens(t *testing.T) {
+	resp := `{
+		"kind": "identitytoolkit#SetAccountInfoResponse",
+		"localId": "expectedUserID"
+	}`
+	s := echoServer([]byte(resp), t)
+	defer s.Close()
+	before := time.Now().Unix()
+	err := s.Client.RevokeRefreshTokens(nil, "some_uid")
+	if err != nil {
```
hiranya911: Combine with the previous line.

auth/user_mgt_test.go
```diff
@@ -496,6 +498,43 @@ func TestUpdateUser(t *testing.T) {
...
 }
+func TestRevokeRefreshTokens(t *testing.T) {
+	resp := `{
+		"kind": "identitytoolkit#SetAccountInfoResponse",
+		"localId": "expectedUserID"
+	}`
+	s := echoServer([]byte(resp), t)
+	defer s.Close()
+	before := time.Now().Unix()
+	err := s.Client.RevokeRefreshTokens(nil, "some_uid")
```
hiranya911: Why are we passing a nil context? And why doesn't that trigger an error?

integration/auth/auth_test.go
```diff
@@ -75,10 +75,65 @@ func TestCustomToken(t *testing.T) {
 	if vt.UID != "user1" {
 		t.Errorf("UID = %q; want UID = %q", vt.UID, "user1")
 	}
+	if err = client.DeleteUser(context.Background(), "user1"); err != nil {
+		t.Error(err)
+	}
+}
+
+func TestCustomTokenVerifyCheckRevoked(t *testing.T) {
```
hiranya911: TestVerifyIDTokenWithCheckRevoked

integration/auth/auth_test.go
```diff
@@ -75,10 +75,65 @@ func TestCustomToken(t *testing.T) {
...
+	}
+	ctx := context.Background()
+	vt, err := client.VerifyIDTokenWithCheckRevoked(ctx, idt)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if vt.UID != uid {
+		t.Errorf("UID = %q; want UID = %q", vt.UID, uid)
+	}
+	time.Sleep(time.Second)
```
hiranya911: Here's where we should be documenting the race condition

integration/auth/auth_test.go
```diff
@@ -75,10 +75,65 @@ func TestCustomToken(t *testing.T) {
...
+	vt, err = client.VerifyIDTokenWithCheckRevoked(ctx, idt)
+	we := "id token has been revoked"
+	if vt != nil || err == nil || err.Error() != we {
+		t.Errorf("tok, err := VerifyIDTokenWithCheckRevoked(); got (%v, %s) ; want (%v, %v)",
+			vt, err, nil, we)
+	}
+
+	// Does not return error if it isn't checked
+	_, err = client.VerifyIDToken(idt)
+	if err != nil {
```
hiranya911: Combine with previous line

integration/auth/auth_test.go
```diff
@@ -75,10 +75,65 @@ func TestCustomToken(t *testing.T) {
...
+	if idt, err = signInWithCustomToken(ct); err != nil {
+		t.Fatal(err)
+	}
+
+	if _, err = client.VerifyIDTokenWithCheckRevoked(ctx, idt); err != nil {
+		t.Errorf("VerifyIDTokenWithCheckRevoked(); err = %s; want err = <nil>", err)
+	}
+
+	err = client.DeleteUser(ctx, uid)
+	if err != nil {
```
hiranya911: Combine with previous line

integration/auth/auth_test.go
```diff
@@ -95,15 +150,20 @@ func TestCustomTokenWithClaims(t *testing.T) {
...
 		t.Errorf("Claims['premium'] = %v; want Claims['premium'] = true", vt.Claims["premium"])
 	}
 	if pkg, ok := vt.Claims["package"].(string); !ok || pkg != "gold" {
 		t.Errorf("Claims['package'] = %v; want Claims['package'] = \"gold\"", vt.Claims["package"])
 	}
+
+	if err = client.DeleteUser(context.Background(), "user2"); err != nil {
+		t.Error(err)
+	}
+
```
hiranya911: Remove empty line

integration/auth/user_mgt_test.go
```diff
@@ -34,16 +36,26 @@ var testFixtures = struct {
...
+		{"Disable user account", testDisableUser},
+		{"Update user", testUpdateUser},
+		{"Remove user attributes", testRemovePhonePhotoName},
+		{"Remove custom claims", testRemoveCustomClaims},
+		{"Add custom claims", testAddCustomClaims},
+		{"Delete test users", testDeleteUsers},
+	}
+	for _, run := range orderedRuns {
+		if ok := t.Run(run.name, run.testFunc); !ok {
+			t.Fatalf("Failed run %v", run.name)
```
hiranya911: Why is this needed? Should add a comment if it's intentional

integration/auth/user_mgt_test.go
```diff
@@ -52,19 +64,33 @@ func TestUserManagement(t *testing.T) {
 func testCreateUsers(t *testing.T) {
 	// Create users with uid
 	for i := 0; i < 3; i++ {
-		params := (&auth.UserToCreate{}).UID(fmt.Sprintf("tempTestUserID-%d", i))
+		uid := fmt.Sprintf("tempTestUserID-%d", i)
+		params := (&auth.UserToCreate{}).UID(uid)
 		u, err := client.CreateUser(context.Background(), params)
 		if err != nil {
-			t.Fatal("failed to create user", i, err)
+			if strings.Contains(err.Error(), "DUPLICATE_LOCAL_ID") {
```
hiranya911: Why?

integration/auth/user_mgt_test.go
```diff
@@ -52,19 +64,33 @@ func TestUserManagement(t *testing.T) {
...
 	// Create user with no parameters (zero-value)
 	u, err := client.CreateUser(context.Background(), (&auth.UserToCreate{}))
 	if err != nil {
 		t.Fatal(err)
 	}
+	// make sure that the user.TokensValidAfterTime is not in the future or stale.
+	if u.TokensValidAfterTime > time.Now().Unix()*1000 {
+		t.Errorf("timestamp cannot be in the future")
+	}
+	if u.TokensValidAfterTime < (time.Now().Unix()-3600)*1000 {
```
hiranya911: If you want: `time.Now().Add(time.Duration(-1) * time.Hour)` will do the date math.

integration/auth/user_mgt_test.go
```diff
@@ -52,19 +64,33 @@ func TestUserManagement(t *testing.T) {
...
+			}
 		}
 		testFixtures.uidList = append(testFixtures.uidList, u.UID)
 	}
-
 	// Create user with no parameters (zero-value)
 	u, err := client.CreateUser(context.Background(), (&auth.UserToCreate{}))
 	if err != nil {
 		t.Fatal(err)
 	}
```
hiranya911: Isn't the default `TokensValidAfterTime` equal to 0? How does the following test pass if that's the case?

integration/auth/user_mgt_test.go
```diff
@@ -76,7 +102,7 @@ func testCreateUsers(t *testing.T) {
 		DisplayName("display_name").
 		Password("password")
 	u, err = client.CreateUser(context.Background(), params)
-	if err != nil {
+	if err != nil || u == nil {
```
hiranya911: Why?

auth/user_mgt.go
avishalom: we are setting the validSince property directly. (from a private function)
validSince is in seconds, TokensValidAfterTime is in millis.

auth/user_mgt_test.go
avishalom: This isn't making a network call.

hiranya911: Doesn't it call the echo server?

auth/user_mgt.go
hiranya911: Then please remove this check.

integration/auth/user_mgt_test.go
avishalom: Adding comment. 
	// The tests are meant to be run in sequence, a failure in creating the users
	// should be fatal so non of the other tests run. However calling Fatal from a
	// subtest does not prevent the other subtests from running, hence we check the
	// success of each subtest before proceeding.

integration/auth/user_mgt_test.go
avishalom: When running the integration tests, if they fail, or you exit them with CTRL-C, residual users on the backend should not prevent the tests from running if we want to create them anyway. 
This checks for the source of failure, and if the user already exists from the previous run, we let it slide.

integration/auth/user_mgt_test.go
avishalom: removed

integration/auth/user_mgt_test.go
avishalom: I was mistaken, 
It defaults to the user creation time.

integration/auth/user_mgt_test.go
avishalom: Thanks! (-time.Hour  also works, but i changed it around a bit. )

CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
 # Unreleased
-- 
 
 ### Token revokaction
 - [added] A New ['VerifyIDTokenWithCheckRevoked(ctx, token)'](https://godoc.org/firebase.google.com/go/auth#Client.VerifyIDToken)
-  method checks to see if the token has been revoked. A new method ['RevokeRefreshTokens(uid)'](https://godoc.org/firebase.google.com/go/auth#Client.RevokeRefreshTokens)
-  has been added to invalidate all tokens issued before the current second. [added] A new property 
+  method checks to see if the token has been revoked. 
+- [added] A new method ['RevokeRefreshTokens(uid)'](https://godoc.org/firebase.google.com/go/auth#Client.RevokeRefreshTokens)
+  has been added to invalidate all tokens issued before the current second.
```
hiranya911: ... all refresh tokens issued to a user.

auth/auth.go
```diff
@@ -177,12 +177,12 @@ func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interfac
 	return encodeToken(c.snr, defaultHeader(), payload)
 }
 
+// RevokeRefreshTokens revokes all refresh tokens for the specified user.
+//
 // RevokeRefreshTokens revokes all refresh tokens for the specified user.
```
hiranya911: This is repeated now.

auth/user_mgt.go
hiranya911: I'd just do a simple sign check to begin with. We can implement more restrictive checks later if it becomes necessary.

integration/auth/auth_test.go
hiranya911: This is missing a full stop somewhere.

integration/auth/user_mgt_test.go
```diff
@@ -51,13 +51,31 @@ func TestUserManagement(t *testing.T) {
 		{"Add custom claims", testAddCustomClaims},
 		{"Delete test users", testDeleteUsers},
 	}
+	// The tests are meant to be run in sequence, a failure in creating the users
```
hiranya911: Replace comma with full stop?

integration/auth/user_mgt_test.go
```diff
@@ -51,13 +51,31 @@ func TestUserManagement(t *testing.T) {
...
+	// subtest does not prevent the other subtests from running, hence we check the
+	// success of each subtest before proceeding.
 	for _, run := range orderedRuns {
 		if ok := t.Run(run.name, run.testFunc); !ok {
 			t.Fatalf("Failed run %v", run.name)
 		}
 	}
 }
 
+func forceCreateUser(uid string, params *auth.UserToCreate) (*auth.UserRecord, error) {
```
hiranya911: This looks hacky. Can we remove this from this PR? We should implement a better solution in a separate PR.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Unreleased
-- 
+
+### Token revokaction
+- [added] A New ['VerifyIDTokenWithCheckRevoked(ctx, token)'](https://godoc.org/firebase.google.com/go/auth#Client.VerifyIDToken)
+  method checks to see if the token has been revoked.
```
hiranya911: method has been added to check for revoked ID tokens.

auth/auth.go
```diff
@@ -177,13 +177,24 @@ func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interfac
 	return encodeToken(c.snr, defaultHeader(), payload)
 }
 
+// RevokeRefreshTokens revokes all refresh tokens for the specified user.
+//
+// RevokeRefreshTokens revokes all refresh tokens for the specified user.
+// In addition to revoking all refresh tokens for a user, all ID tokens issued
+// before revocation will also be revoked at the Auth backend. Any request with an
```
hiranya911: Hmmm... I don't think this is correct. Can we align this with what's documented at https://firebase.google.com/docs/reference/admin/node/admin.auth.Auth#revokeRefreshTokens ?

auth/auth.go
```diff
@@ -177,13 +177,24 @@ func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interfac
...
+}
+
 // VerifyIDToken verifies the signature	and payload of the provided ID token.
 //
 // VerifyIDToken accepts a signed JWT token string, and verifies that it is current, issued for the
 // correct Firebase project, and signed by the Google Firebase services in the cloud. It returns
 // a Token containing the decoded claims in the input JWT. See
 // https://firebase.google.com/docs/auth/admin/verify-id-tokens#retrieve_id_tokens_on_clients for
 // more details on how to obtain an ID token in a client app.
+// This does not check whether or not the token has been revoked, see VerifyIDTokenWithCheckRevoked below.
```
hiranya911: Comma --> Full stop

auth/user_mgt.go
avishalom: It is valid since internally. it is exposed as tokensValidAfterMillis

auth/auth.go
avishalom: Correction, valid since is initialized to the current timestamp.

auth/auth.go
```diff
@@ -179,10 +179,12 @@ func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interfac
 
 // RevokeRefreshTokens revokes all refresh tokens for the specified user.
 //
-// RevokeRefreshTokens revokes all refresh tokens for the specified user.
-// In addition to revoking all refresh tokens for a user, all ID tokens issued
-// before revocation will also be revoked at the Auth backend. Any request with an
-// ID token generated before revocation will be rejected with a token expired error.
+// RevokeRefreshTokens updates the user's TokensValidAfterMillis to the current UTC second.
```
hiranya911: Please document this from the developer's point of view, and not SDK's point of view. Something like:

```
 RevokeRefreshTokens revokes all refresh tokens issued to a user.

...
```

auth/auth.go
```diff
@@ -194,7 +196,7 @@ func (c *Client) RevokeRefreshTokens(ctx context.Context, uid string) error {
 // a Token containing the decoded claims in the input JWT. See
 // https://firebase.google.com/docs/auth/admin/verify-id-tokens#retrieve_id_tokens_on_clients for
 // more details on how to obtain an ID token in a client app.
-// This does not check whether or not the token has been revoked, see VerifyIDTokenAndCheckRevoked below.
+// This does not check whether or not the token has been revoked. see `VerifyIDTokenAndCheckRevoked` below.
```
hiranya911: See (capitalization)

auth/user_mgt.go
```diff
@@ -428,8 +428,8 @@ func validatePhone(val interface{}) error {
 }
 
 func validateValidSince(val interface{}) error {
-	if _, ok := val.(int64); !ok {
-		return fmt.Errorf("validSince must be an integer signifying seconds since the epoch")
+	if v, ok := val.(int64); !ok || v <= 0 {
+		return fmt.Errorf("validSince must be an positive integer signifying seconds since the epoch")
```
hiranya911: a positive integer representing...

auth/user_mgt.go
```diff
@@ -416,6 +427,13 @@ func validatePhone(val interface{}) error {
 	return nil
 }
 
+func validateValidSince(val interface{}) error {
```
hiranya911: Given that there's no way for a user to set this explicitly, I wonder if this check is even needed. Consider removing.

auth/auth.go
avishalom: This is pretty much in line with the Node docs. I am not sure I understand the comment. (i change the first line as suggested.)

auth/auth.go
```diff
@@ -177,13 +177,26 @@ func (c *Client) CustomTokenWithClaims(uid string, devClaims map[string]interfac
 	return encodeToken(c.snr, defaultHeader(), payload)
 }
 
+// RevokeRefreshTokens revokes all refresh tokens issued to a user.
+//
+// RevokeRefreshTokens updates the user's TokensValidAfterMillis to the current UTC second.
```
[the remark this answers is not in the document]
bklimt: Oh, I guess second is correct. I misread it before.

PR #62, opened 2018-01-02
Adding Firebase Cloud Messaging

chemidy:
I just add [Firebase Cloud Messaging V1 API](https://firebase.google.com/docs/cloud-messaging/send-message) support

But it's not very easy to test everything with integration test (all kind : APNS, Android, WebPush with different configurations : data, notification).

PR #81, opened 2018-02-01
Firebase Cloud Messaging API

hiranya911:
This PR builds on the work by @chemidy at #62. Implements the following public API calls in the new `messaging` package:

```
Send(ctx context.Context, m *Message) (string, error)
SubscribeToTopic(ctx context.Context, tokens []string, topic string) (*TopicManagementResponse, error)
UnsubscribeFromTopic(ctx context.Context, tokens []string, topic string) (*TopicManagementResponse, error)
```

go/firebase-admin-fcm-api

### DIFF
--- a/integration/messaging/messaging_test.go
+++ b/integration/messaging/messaging_test.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"regexp"
 	"testing"
-	"time"
 
 	"golang.org/x/net/context"
 
@@ -34,15 +33,7 @@ import (
 const testRegistrationToken = "fGw0qy4TGgk:APA91bGtWGjuhp4WRhHXgbabIYp1jxEKI08ofj_v1bKhWAGJQ4e3a" +
 	"rRCWzeTfHaLz83mBnDh0aPWB1AykXAVUUGl2h1wT4XI6XazWpvY7RBUSYfoxtqSWGIm2nvWh2BOP1YG501SsRoE"
 
-var (
-	client       *messaging.Client
-	testFixtures = struct {
-		token     string
-		topic     string
-		condition string
-	}{}
-	ttl = time.Duration(3) * time.Second
-)
+var client *messaging.Client
 
 // Enable API before testing
 // https://console.developers.google.com/apis/library/fcm.googleapis.com
--- a/integration/messaging/messaging_test.go
+++ b/integration/messaging/messaging_test.go
@@ -15,7 +15,6 @@
 package messaging
 
 import (
-	"context"
 	"flag"
 	"log"
 	"os"
@@ -23,9 +22,14 @@ import (
 	"testing"
 	"time"
 
+	"golang.org/x/net/context"
+
 	"firebase.google.com/go/integration/internal"
 	"firebase.google.com/go/messaging"
 )
+
+const testRegistrationToken = "fGw0qy4TGgk:APA91bGtWGjuhp4WRhHXgbabIYp1jxEKI08ofj_v1bKhWAGJQ4e3arRCWzeTfHaLz83mBnDh0a" +
+	"PWB1AykXAVUUGl2h1wT4XI6XazWpvY7RBUSYfoxtqSWGIm2nvWh2BOP1YG501SsRoE"
 
 var client *messaging.Client
 
@@ -60,7 +64,6 @@ func TestMain(m *testing.M) {
 }
 
 func TestSend(t *testing.T) {
-	ctx := context.Background()
 	msg := &messaging.Message{
 		Topic: "foo-bar",
 		Notification: &messaging.Notification{
@@ -90,7 +93,7 @@ func TestSend(t *testing.T) {
 			},
 		},
 	}
-	name, err := client.SendDryRun(ctx, msg)
+	name, err := client.SendDryRun(context.Background(), msg)
 	if err != nil {
 		log.Fatal(err)
 	}
@@ -101,10 +104,29 @@ func TestSend(t *testing.T) {
 }
 
 func TestSendInvalidToken(t *testing.T) {
-	ctx := context.Background()
 	msg := &messaging.Message{Token: "INVALID_TOKEN"}
-	_, err := client.Send(ctx, msg)
+	_, err := client.Send(context.Background(), msg)
 	if err == nil {
 		t.Errorf("Send() = nil; want error")
+	}
+}
+
+func TestSubscribe(t *testing.T) {
+	tmr, err := client.SubscribeToTopic(context.Background(), []string{testRegistrationToken}, "mock-topic")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if tmr.SuccessCount+tmr.FailureCount != 1 {
+		t.Errorf("SubscribeToTopic() = %v; want total 1", tmr)
+	}
+}
+
+func TestUnsubscribe(t *testing.T) {
+	tmr, err := client.UnsubscribeFromTopic(context.Background(), []string{testRegistrationToken}, "mock-topic")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if tmr.SuccessCount+tmr.FailureCount != 1 {
+		t.Errorf("UnsubscribeFromTopic() = %v; want total 1", tmr)
 	}
 }
--- a/integration/messaging/messaging_test.go
+++ b/integration/messaging/messaging_test.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"testing"
+	"time"
 
 	"firebase.google.com/go/integration/internal"
 	"firebase.google.com/go/messaging"
@@ -21,6 +22,8 @@ var testFixtures = struct {
 	topic     string
 	condition string
 }{}
+
+var ttl = time.Duration(3) * time.Second
 
 // Enable API before testing
 // https://console.developers.google.com/apis/library/fcm.googleapis.com/?project=
@@ -188,7 +191,7 @@ func TestSendData(t *testing.T) {
 	ctx := context.Background()
 	msg := &messaging.Message{
 		Token: testFixtures.token,
-		Data: map[string]interface{}{
+		Data: map[string]string{
 			"private_key":  "foo",
 			"client_email": "bar@test.com",
 		},
@@ -215,7 +218,7 @@ func TestSendAndroidNotification(t *testing.T) {
 		Android: &messaging.AndroidConfig{
 			CollapseKey: "Collapse",
 			Priority:    "HIGH",
-			TTL:         "3.5s",
+			TTL:         &ttl,
 			Notification: &messaging.AndroidNotification{
 				Title: "Android Title",
 				Body:  "Android body",
@@ -244,7 +247,7 @@ func TestSendAndroidData(t *testing.T) {
 		Android: &messaging.AndroidConfig{
 			CollapseKey: "Collapse",
 			Priority:    "HIGH",
-			TTL:         "3.5s",
+			TTL:         &ttl,
 			Data: map[string]string{
 				"private_key":  "foo",
 				"client_email": "bar@test.com",
@@ -271,9 +274,13 @@ func TestSendAPNSNotification(t *testing.T) {
 			Body:  "This is a Notification",
 		},
 		APNS: &messaging.APNSConfig{
-			Payload: map[string]string{
-				"title": "APNS Title ",
-				"body":  "APNS bodym",
+			Payload: &messaging.APNSPayload{
+				Aps: &messaging.Aps{
+					Alert: &messaging.ApsAlert{
+						Title: "APNS title",
+						Body:  "APNS body",
+					},
+				},
 			},
 		},
 	}
--- a/integration/messaging/messaging_test.go
+++ b/integration/messaging/messaging_test.go
@@ -28,25 +28,28 @@ import (
 	"firebase.google.com/go/messaging"
 )
 
-const testRegistrationToken = "fGw0qy4TGgk:APA91bGtWGjuhp4WRhHXgbabIYp1jxEKI08ofj_v1bKhWAGJQ4e3arRCWzeTfHaLz83mBnDh0a" +
-	"PWB1AykXAVUUGl2h1wT4XI6XazWpvY7RBUSYfoxtqSWGIm2nvWh2BOP1YG501SsRoE"
+// The registration token has the proper format, but is not valid (i.e. expired). The intention of
+// these integration tests is to verify that the endpoints return the proper payload, but it is
+// hard to ensure this token remains valid. The tests below should still pass regardless.
+const testRegistrationToken = "fGw0qy4TGgk:APA91bGtWGjuhp4WRhHXgbabIYp1jxEKI08ofj_v1bKhWAGJQ4e3a" +
+	"rRCWzeTfHaLz83mBnDh0aPWB1AykXAVUUGl2h1wT4XI6XazWpvY7RBUSYfoxtqSWGIm2nvWh2BOP1YG501SsRoE"
 
-var client *messaging.Client
-
-var testFixtures = struct {
-	token     string
-	topic     string
-	condition string
-}{}
-
-var ttl = time.Duration(3) * time.Second
+var (
+	client       *messaging.Client
+	testFixtures = struct {
+		token     string
+		topic     string
+		condition string
+	}{}
+	ttl = time.Duration(3) * time.Second
+)
 
 // Enable API before testing
-// https://console.developers.google.com/apis/library/fcm.googleapis.com/?project=
+// https://console.developers.google.com/apis/library/fcm.googleapis.com
 func TestMain(m *testing.M) {
 	flag.Parse()
 	if testing.Short() {
-		log.Println("skipping Messaging integration tests in short mode.")
+		log.Println("Skipping messaging integration tests in short mode.")
 		return
 	}
 
@@ -95,7 +98,7 @@ func TestSend(t *testing.T) {
 	}
 	name, err := client.SendDryRun(context.Background(), msg)
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalln(err)
 	}
 	const pattern = "^projects/.*/messages/.*$"
 	if !regexp.MustCompile(pattern).MatchString(name) {
@@ -105,8 +108,7 @@ func TestSend(t *testing.T) {
 
 func TestSendInvalidToken(t *testing.T) {
 	msg := &messaging.Message{Token: "INVALID_TOKEN"}
-	_, err := client.Send(context.Background(), msg)
-	if err == nil {
+	if _, err := client.Send(context.Background(), msg); err == nil {
 		t.Errorf("Send() = nil; want error")
 	}
 }
--- a/integration/messaging/messaging_test.go
+++ b/integration/messaging/messaging_test.go
@@ -1,0 +1,315 @@
+package messaging
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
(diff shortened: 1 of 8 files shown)


integration/messaging/messaging_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"testing"
+	"time"
+
+	"golang.org/x/net/context"
+
+	"firebase.google.com/go/integration/internal"
+	"firebase.google.com/go/messaging"
+)
+
+const testRegistrationToken = "fGw0qy4TGgk:APA91bGtWGjuhp4WRhHXgbabIYp1jxEKI08ofj_v1bKhWAGJQ4e3arRCWzeTfHaLz83mBnDh0a" +
```
bklimt: Please add a comment indicating this is not used.

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+}
+
+func newTopicManagementResponse(resp *iidResponse) *TopicManagementResponse {
+	tmr := &TopicManagementResponse{}
+	for idx, res := range resp.Results {
+		if len(res) == 0 {
+			tmr.SuccessCount++
+		} else {
+			tmr.FailureCount++
+			code, _ := res["error"].(string)
```
sampson-chen: If there's a rationale for explicitly ignoring the error from the type assertion, let's document it. 

If the rationale is "// should never happen", we can just omit the _ and let it panic if a non-string is found.

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+	return tmr
+}
+
+// NewClient creates a new instance of the Firebase Cloud Messaging Client.
+//
+// This function can only be invoked from within the SDK. Client applications should access the
+// the messaging service through firebase.App.
+func NewClient(ctx context.Context, c *internal.MessagingConfig) (*Client, error) {
+	if c.ProjectID == "" {
+		return nil, errors.New("project id is required to access firebase cloud messaging client")
```
sampson-chen: s/project id is required to access firebase cloud messaging client/project ID is required to access Firebase Cloud Messaging client/

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+		client:      &internal.HTTPClient{Client: hc},
+		project:     c.ProjectID,
+		version:     "Go/Admin/" + c.Version,
+	}, nil
+}
+
+// Send sends a Message to Firebase Cloud Messaging.
+//
+// The Message must specify exactly one of Token, Topic and Condition fields. FCM will
+// customize the message for each target platform based on the parameters specified in the
```
sampson-chen: Super pedantic nit: https://stackoverflow.com/questions/156767/whats-the-difference-between-an-argument-and-a-parameter

Fine as-is.

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+		var result fcmResponse
+		err := json.Unmarshal(resp.Body, &result)
+		return result.Name, err
+	}
+
+	var fe fcmError
+	json.Unmarshal(resp.Body, &fe) // ignore any json parse errors at this level
+	msg := fcmErrorCodes[fe.Error.Status]
+	if msg == "" {
+		msg = fmt.Sprintf("client encounterd an unknown error; response: %s", string(resp.Body))
```
sampson-chen: s/encounterd/encountered/

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+	for _, token := range req.Tokens {
+		if token == "" {
+			return nil, fmt.Errorf("tokens list must not contain empty strings")
+		}
+	}
+
+	if req.Topic == "" {
+		return nil, fmt.Errorf("topic name not specified")
+	}
+	if !regexp.MustCompile("^(/topics/)?(private/)?[a-zA-Z0-9-_.~%]+$").MatchString(req.Topic) {
```
sampson-chen: How about pulling this into a var so we don't re-run MustCompile every time this is called?

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+			return nil, err
+		}
+		return newTopicManagementResponse(&result), nil
+	}
+
+	var ie iidError
+	json.Unmarshal(resp.Body, &ie) // ignore any json parse errors at this level
+	msg := iidErrorCodes[ie.Error]
+	if msg == "" {
+		msg = fmt.Sprintf("client encounterd an unknown error; response: %s", string(resp.Body))
```
sampson-chen: s/encounterd/encountered/

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+	"PERMISSION_DENIED":  "client does not have permission to perform the requested operation; code: authentication-error",
+	"RESOURCE_EXHAUSTED": "messaging service quota exceeded; code: message-rate-exceeded",
+	"UNAUTHENTICATED":    "client failed to authenticate; code: authentication-error",
+	"UNAVAILABLE":        "backend servers are temporarily unavailable; code: server-unavailable",
+}
+
+var iidErrorCodes = map[string]string{
+	"INVALID_ARGUMENT": "request contains an invalid argument; code: invalid-argument",
+	"NOT_FOUND":        "request contains an invalid argument; code: registration-token-not-registered",
+	"INTERNAL":         "server encounterd an internal error; code: internal-error",
```
sampson-chen: s/encounterd/encountered/

messaging/messaging_test.go
```diff
@@ -0,0 +1,829 @@
...
+	}
+	client.fcmEndpoint = ts.URL
+
+	cases := []struct {
+		resp string
+		want string
+	}{
+		{
+			resp: "{}",
+			want: "http error status: 500; reason: client encounterd an unknown error; response: {}",
```
sampson-chen: s/encounterd/encountered/, here and elsewhere

messaging/messaging_utils.go
```diff
@@ -0,0 +1,126 @@
...
+	if target != 1 {
+		return fmt.Errorf("exactly one of token, topic or condition must be specified")
+	}
+
+	// validate topic
+	if message.Topic != "" {
+		if strings.HasPrefix(message.Topic, "/topics/") {
+			return fmt.Errorf("topic name must not contain the /topics/ prefix")
+		}
+		if !regexp.MustCompile("^[a-zA-Z0-9-_.~%]+$").MatchString(message.Topic) {
```
sampson-chen: Similarly here and elsewhere, maybe pull out the MustCompile to a var

messaging/messaging_test.go
```diff
@@ -0,0 +1,829 @@
...
+		name:   "TooManyTokens",
+		tokens: strings.Split("a"+strings.Repeat(",a", 1000), ","),
+		topic:  "topic",
+		want:   "tokens list must not contain more than 1000 items",
+	},
+	{
+		name:   "EmptyToken",
+		tokens: []string{"foo", ""},
+		topic:  "topic",
+		want:   "tokens list must not contain empty strings",
```
sampson-chen: I've been told previously by other Go readability reviewers that string comparison on error messages is brittle and thus discouraged, but I'm having trouble finding a referenceable link that addresses this issue and what they suggest otherwise.

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+	payload := &fcmRequest{
+		ValidateOnly: true,
+		Message:      message,
+	}
+	return c.makeSendRequest(ctx, payload)
+}
+
+// SubscribeToTopic subscribes a list of registration tokens to a topic.
+//
+// The tokens list must not be empty, with at most 1000 tokens.
```
avishalom: nit: s/, with at most/ and must have at most/
(or something like that...)

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+		msg = fmt.Sprintf("client encounterd an unknown error; response: %s", string(resp.Body))
+	}
+	return "", fmt.Errorf("http error status: %d; reason: %s", resp.Status, msg)
+}
+
+func (c *Client) makeTopicManagementRequest(ctx context.Context, req *iidRequest) (*TopicManagementResponse, error) {
+	if req.Tokens == nil {
+		return nil, fmt.Errorf("no tokens specified")
+	}
+	if len(req.Tokens) > 1000 {
```
avishalom: Are we also testing len() > 0. (as suggested in the comment.)

hiranya911: This is same as `req.Tokens == nil`

messaging/messaging.go

messaging/messaging_utils.go
hiranya911: Done

messaging/messaging_test.go
hiranya911: Done

messaging/messaging_test.go
hiranya911: I agree in general. However, at least in this case we are testing for the errors produced by our own code.

messaging/messaging.go

messaging/messaging.go
hiranya911: Done, good idea.

messaging/messaging.go

messaging/messaging.go
hiranya911: Changed to arguments

messaging/messaging.go

messaging/messaging.go
hiranya911: I suppose if we've made this far into the error parsing process, we can assume the `error` key is present and contains a string value. Dropping the `_`.

messaging/messaging.go
```diff
@@ -0,0 +1,454 @@
...
+	Condition    string            `json:"condition,omitempty"`
+}
+
+// Notification is the basic notification template to use across all platforms.
+type Notification struct {
+	Title string `json:"title,omitempty"`
+	Body  string `json:"body,omitempty"`
+}
+
+// AndroidConfig contains messaging options specific to the Android platform..
```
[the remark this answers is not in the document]

messaging/messaging.go

integration/messaging/messaging_test.go
```diff
@@ -0,0 +1,132 @@
...
+	const pattern = "^projects/.*/messages/.*$"
+	if !regexp.MustCompile(pattern).MatchString(name) {
+		t.Errorf("Send() = %q; want = %q", name, pattern)
+	}
+}
+
+func TestSendInvalidToken(t *testing.T) {
+	msg := &messaging.Message{Token: "INVALID_TOKEN"}
+	_, err := client.Send(context.Background(), msg)
+	if err == nil {
```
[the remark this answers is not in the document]
hiranya911: Done

integration/messaging/messaging_test.go
```diff
@@ -0,0 +1,132 @@
...
+		Webpush: &messaging.WebpushConfig{
+			Notification: &messaging.WebpushNotification{
+				Title: "Webpush Title",
+				Body:  "Webpush Body",
+			},
+		},
+	}
+	name, err := client.SendDryRun(context.Background(), msg)
+	if err != nil {
+		log.Fatal(err)
```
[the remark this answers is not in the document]
hiranya911: Changed to `Fatalln`. Thanks for pointing it out.

integration/messaging/messaging_test.go
```diff
@@ -0,0 +1,132 @@
...
+}{}
+
+var ttl = time.Duration(3) * time.Second
+
+// Enable API before testing
+// https://console.developers.google.com/apis/library/fcm.googleapis.com/?project=
+func TestMain(m *testing.M) {
+	flag.Parse()
+	if testing.Short() {
+		log.Println("skipping Messaging integration tests in short mode.")
```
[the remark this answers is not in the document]
hiranya911: Done (we should do the same in other packages eventually)

integration/messaging/messaging_test.go
```diff
@@ -0,0 +1,132 @@
...
+var testFixtures = struct {
+	token     string
+	topic     string
+	condition string
+}{}
+
+var ttl = time.Duration(3) * time.Second
+
+// Enable API before testing
+// https://console.developers.google.com/apis/library/fcm.googleapis.com/?project=
```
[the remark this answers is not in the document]
hiranya911: Done

integration/messaging/messaging_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"golang.org/x/net/context"
+
+	"firebase.google.com/go/integration/internal"
+	"firebase.google.com/go/messaging"
+)
+
+const testRegistrationToken = "fGw0qy4TGgk:APA91bGtWGjuhp4WRhHXgbabIYp1jxEKI08ofj_v1bKhWAGJQ4e3arRCWzeTfHaLz83mBnDh0a" +
+	"PWB1AykXAVUUGl2h1wT4XI6XazWpvY7RBUSYfoxtqSWGIm2nvWh2BOP1YG501SsRoE"
+
+var client *messaging.Client
```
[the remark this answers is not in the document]
hiranya911: Done

integration/messaging/messaging_test.go
hiranya911: Done

hiranya911:
Addressed all open comments up to this point. Over to you for another look.

messaging/messaging_test.go
sampson-chen: Good point. SG.

messaging/messaging.go
sampson-chen: This is a good point. `len(req.Tokens) == 0` is a more defensive check than `req.Tokens == nil` because it's possible to have a non-nil zero-length slice: https://play.golang.org/p/IPdRvwsFutM

messaging/messaging_utils.go
```diff
@@ -0,0 +1,130 @@
...
+	bareTopicNamePattern = regexp.MustCompile("^[a-zA-Z0-9-_.~%]+$")
+	colorPattern         = regexp.MustCompile("^#[0-9a-fA-F]{6}$")
+)
+
+func validateMessage(message *Message) error {
+	if message == nil {
+		return fmt.Errorf("message must not be nil")
+	}
+
+	target := bool2int(message.Token != "") + bool2int(message.Condition != "") + bool2int(message.Topic != "")
```
avishalom: I think that 
```
target := 0
for _, v := range []bool{message.Token != "", message.Condition != "", message.Topic != ""} {
  if v {
    target +=1
}
```

is slightly more readable, I don't feel very strongly about this though. 

maybe make a function of it at a slightly higher level than the single translation

```
func countTrues(arr []bool) {
  res := 0 
    for _, v := range arr {
      if v {
        res += 1
      }
  }
 return res
}
```
or better yet
```
func countNonEmpty(arr []string) {
  res := 0 
    for _, s := range arr {
      if s != "" {
        res += 1
      }
  }
 return res
}

File: messaging/messaging.go (hunk 13 above)
Comment (hiranya911) [reply messaging/messaging.go]: Ah, makes sense. Making the change.

File: messaging/messaging_utils.go (hunk 20 above)
Comment (hiranya911) [reply messaging/messaging_utils.go]: Good idea. Done

**Comment by hiranya911** [comment]
@bklimt I'll wait for your feedback to finish this off.

File: integration/messaging/messaging_test.go [hunk 21]
```diff
@@ -0,0 +1,134 @@
...
+
+// The registration token has the proper format, but is not valid (i.e. expired). The intention of
+// these integration tests is to verify that the endpoints return the proper payload, but it is
+// hard to ensure this token remains valid. The tests below should still pass regardless.
+const testRegistrationToken = "fGw0qy4TGgk:APA91bGtWGjuhp4WRhHXgbabIYp1jxEKI08ofj_v1bKhWAGJQ4e3a" +
+	"rRCWzeTfHaLz83mBnDh0aPWB1AykXAVUUGl2h1wT4XI6XazWpvY7RBUSYfoxtqSWGIm2nvWh2BOP1YG501SsRoE"
+
+var (
+	client       *messaging.Client
+	testFixtures = struct {
```
bklimt: It looks like this isn't used?

integration/messaging/messaging_test.go
```diff
@@ -0,0 +1,134 @@
...
+		token     string
+		topic     string
+		condition string
+	}{}
+	ttl = time.Duration(3) * time.Second
```
bklimt: Is this used?

messaging/messaging.go
```diff
@@ -0,0 +1,460 @@
...
+	messagingEndpoint = "https://fcm.googleapis.com/v1"
+	iidEndpoint       = "https://iid.googleapis.com"
+	iidSubscribe      = "iid/v1:batchAdd"
+	iidUnsubscribe    = "iid/v1:batchRemove"
+)
+
+var (
+	topicNamePattern = regexp.MustCompile("^(/topics/)?(private/)?[a-zA-Z0-9-_.~%]+$")
+
+	fcmErrorCodes = map[string]string{
```
bklimt: These are mostly the standard Google error codes. Can we make one map that we share across all the SDKs for the standard error codes?

messaging/messaging.go
```diff
@@ -0,0 +1,460 @@
...
+		"RESOURCE_EXHAUSTED": "messaging service quota exceeded; code: message-rate-exceeded",
+		"UNAUTHENTICATED":    "client failed to authenticate; code: authentication-error",
+		"UNAVAILABLE":        "backend servers are temporarily unavailable; code: server-unavailable",
+	}
+
+	iidErrorCodes = map[string]string{
+		"INVALID_ARGUMENT": "request contains an invalid argument; code: invalid-argument",
+		"NOT_FOUND":        "request contains an invalid argument; code: registration-token-not-registered",
+		"INTERNAL":         "server encountered an internal error; code: internal-error",
+		"TOO_MANY_TOPICS":  "client exceeded the number of allowed topics; code: too-many-topics",
```
bklimt: Oh no, iid can have a non-standard error value? Hmm...

integration/messaging/messaging_test.go
hiranya911: Removed.

integration/messaging/messaging_test.go
hiranya911: Removed

messaging/messaging.go
hiranya911: It is documented here: https://developers.google.com/instance-id/reference/server#manage_relationship_maps_for_multiple_app_instances

messaging/messaging.go
hiranya911: We seem to have a few services with non-standard error codes. Also the interpretation of some error codes seem to be different from service to service (e.g. `NOT_FOUND` means one thing in IID, and another in FCM). Lets have a separate discussion about this error code reconciliation.

michaljemala: Error codes enumerated in this map does not match those mentioned in the [docs](https://firebase.google.com/docs/reference/fcm/rest/v1/ErrorCode). Is the documentation up to date?

Also, should `Send` fail would it make sense to expose the error code more explicitly? Right now it is hidden in a string message.

hiranya911: We will start with the bare minimal API surface possible. We can add things like error codes (or possibly package-level error constants as is typical in Go), in the future depending on how developers use it. If we start with a large API surface now, it is going to be difficult to make changes without introducing breaking changes.

michaljemala: I understand it, although it will make it much harder to process the errors programatically.
Moreover considering the other Admin SDKs (Python, Java, Node) will expose the error code, it would make sense to put Go SDK on par with the others.

Also, are the error codes in the docs obsolete?

hiranya911: Sync'ed with the new error codes (documentation is up-to-date). We are still not handling some undocumented error codes, but we can deal with them later.

PR #84, opened 2018-02-07
Accepting prefixed topic names

hiranya911:
The new FCM API does not accept topic names prefixed with `/topics/`. But we can easily check for this in the SDK and remove it. This results in a better and consistent dev experience since the other methods in this API accepts prefixed topic names.

### DIFF
--- a/messaging/messaging.go
+++ b/messaging/messaging.go
@@ -81,8 +81,21 @@ type Message struct {
 	Webpush      *WebpushConfig    `json:"webpush,omitempty"`
 	APNS         *APNSConfig       `json:"apns,omitempty"`
 	Token        string            `json:"token,omitempty"`
-	Topic        string            `json:"topic,omitempty"`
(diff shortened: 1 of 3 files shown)


messaging/messaging.go
```diff
@@ -81,10 +81,23 @@ type Message struct {
...
 	APNS         *APNSConfig       `json:"apns,omitempty"`
 	Token        string            `json:"token,omitempty"`
-	Topic        string            `json:"topic,omitempty"`
+	Topic        string            `json:"-"`
 	Condition    string            `json:"condition,omitempty"`
 }
 
+// MarshalJSON marshals a Message into JSON (for internal use only).
+func (m *Message) MarshalJSON() ([]byte, error) {
+	type messageInternal Message
```
bklimt: I'd add a comment explaining that you have to create a new type to prevent this function from infinitely recursing, assuming that's why it's here. It wasn't obvious.

PR #85, opened 2018-02-07
Document non existing name for iid delete integration test

avishalom:


integration/iid/iid_test.go
```diff
@@ -50,12 +50,13 @@ func TestMain(m *testing.M) {
 }
 
 func TestNonExisting(t *testing.T) {
-	err := client.DeleteInstanceID(context.Background(), "dnon-existY")
+	// legal instance IDs are /[cdef][A-Za-z0-9_-]{9}[AEIMQUYcgkosw048]/
+	err := client.DeleteInstanceID(context.Background(), "fictive-ID0")
```
bklimt: TIL the word  "fictive".

avishalom: It would have worked just as well with a made up word :-)

PR #86, opened 2018-02-07
Import "context" from standard package

avishalom:


hiranya911:
Yes, this is going to terminate our support for older Go versions. Need to have a separate discussion before we do that.

PR #88, opened 2018-02-08
Breaking context change

avishalom:
do not merge, this is a breaking change.
adding context to network calls.

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -139,7 +139,7 @@ func TestCustomTokenWithClaims(t *testing.T) {
 		"premium": true,
 		"count":   float64(123),
 	}
-	token, err := client.CustomTokenWithClaims("user1", claims)
+	token, err := client.CustomTokenWithClaims(ctx, "user1", claims)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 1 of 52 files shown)


integration/iid/iid_test.go
```diff
@@ -51,11 +51,12 @@ func TestMain(m *testing.M) {
 }
 
 func TestNonExisting(t *testing.T) {
-	err := client.DeleteInstanceID(context.Background(), "dnon-existY")
+	// legal instance IDs are /[cdef][A-Za-z0-9_-]{9}[AEIMQUYcgkosw048]/
```
hiranya911: This doesn't look like it belong here. May be it will go away when merged with the latest dev?

auth/auth_test.go
```diff
@@ -37,6 +37,7 @@ import (
 )
 
 var client *Client
+var ctx context.Context
```
hiranya911: Is this ever assigned to? Personally, I think we are better off calling `context.Background()` everywhere this is needed. Otherwise it is somewhat confusing since this is referenced from multiple test files.

avishalom: This gets set in test main for the app engine in line 60, else 68. 
are we dropping the AE support?

integration/iid/iid_test.go
avishalom: this explains fictive-ID0 , i'll change the comment

auth/auth_test.go
hiranya911: Ah I see. Then lets leave it as is for now. Going forward though I think we should remove the dev app server test mode. It's complicating the code here.

PR #93, opened 2018-02-16
Moving the snippets to the main repo.

avishalom:


samtstern:
@avishalom well that was fast!  Before I start reading the code, is this just a straight copy-and-paste from snippets-go?

avishalom:
Yes, it is actually a very slow copy and paste :-)

samtstern:
@avishalom I think we're missing something in the `.travis.yml` that makes sure these compile?

avishalom:
go test -v -test.short ./...
builds everything, (i tried inserting errors. but couldn't get them to pass...)

PR #95, opened 2018-02-17
Refactor snippet files.

avishalom:


hiranya911:
Lets remove the placeholder `main*()` functions. And also lets merge token.go and users.go into a single auth.go file. I think we can establish a convention where each subpackage (auth, messaging, db) has its own snippets file.

PR #102, opened 2018-02-27
Database Sample Snippets

hiranya911:
Snippets to appear at:

* https://firebase.google.com/docs/database/admin/start
* https://firebase.google.com/docs/database/admin/save-data
* https://firebase.google.com/docs/database/admin/retrieve-data

Also made some updates to existing auth snippets to make the use of context more consistent.


samtstern:
All of the `Fatalln` calls have `%v` in them which is causing the build to fail.

PR #109, opened 2018-03-09
Quoted error messages

avishalom:
Fix old nit.

### DIFF
--- a/internal/http_client.go
+++ b/internal/http_client.go
@@ -140,7 +140,7 @@ func (r *Response) CheckStatus(want int) error {
 	if msg == "" {
 		msg = string(r.Body)
 	}
-	return fmt.Errorf("http error status: %d; reason: %s", r.Status, msg)
+	return fmt.Errorf("http error status: %d; reason: %q", r.Status, msg)
 }
 
 // Unmarshal checks if the Response has the given HTTP status code, and if so unmarshals the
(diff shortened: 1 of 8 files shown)


internal/http_client.go
```diff
@@ -140,7 +140,7 @@ func (r *Response) CheckStatus(want int) error {
 	if msg == "" {
 		msg = string(r.Body)
 	}
-	return fmt.Errorf("http error status: %d; reason: %s", r.Status, msg)
+	return fmt.Errorf("http error status: %d; reason: %q", r.Status, msg)
```
hiranya911: This should not be quoted. It is breaking a bunch of tests now.

messaging/messaging.go
```diff
@@ -423,9 +423,9 @@ func (c *Client) makeSendRequest(ctx context.Context, req *fcmRequest) (string,
 	json.Unmarshal(resp.Body, &fe) // ignore any json parse errors at this level
 	msg := fcmErrorCodes[fe.Error.Status]
 	if msg == "" {
-		msg = fmt.Sprintf("server responded with an unknown error; response: %s", string(resp.Body))
+		msg = fmt.Sprintf("server responded with an unknown error; response: %q", string(resp.Body))
```
hiranya911: None of these should be quoted.

internal/http_client.go
avishalom: yes , sorry I  was playing around with it, i though i removed you from review

hiranya911:
FWIW, I didn't object quoting the printed/logged error messages. Only the errors returned from an API.

PR #113, opened 2018-03-11
error message

avishalom:
Fixing the error message for get user by - for non existing user.

### DIFF
--- a/integration/auth/user_mgt_test.go
+++ b/integration/auth/user_mgt_test.go
@@ -414,27 +414,3 @@ func testDeleteUsers(t *testing.T) {
 		}
 	}
 }
-
-func TestGetUserInvalidId(t *testing.T) {
-	we := "cannot find user given params: id:[idbad], phone:[], email: []"
-	_, err := client.GetUser(context.Background(), "idbad")
-	if err == nil || err.Error() != we {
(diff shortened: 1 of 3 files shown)


integration/auth/user_mgt_test.go
```diff
@@ -411,3 +414,27 @@ func testDeleteUsers(t *testing.T) {
 		}
 	}
 }
+
+func TestGetUserInvalidId(t *testing.T) {
```
hiranya911: Suggestion: This sort of logic is easier/cheaper to verify in unit tests. Also these tests would fail if somebody creates a user with this email/phone number in the project. Consider just updating https://github.com/firebase/firebase-admin-go/blob/master/auth/user_mgt_test.go#L143 and removing these.

integration/auth/user_mgt_test.go
```diff
@@ -411,3 +414,27 @@ func testDeleteUsers(t *testing.T) {
...
+	we := "cannot find user given params: id:[idbad], phone:[], email: []"
+	_, err := client.GetUser(context.Background(), "idbad")
```
hiranya911: s/idbad/nonexisting/ for clarity

PR #114, opened 2018-03-12
Fix invalid endpoint URL for topic unsubscribe

michaljemala:
#115 - Uses correct `op` when creating topic unsubscribe requests.


michaljemala:
Just signed the CLA.

PR #126, opened 2018-03-22
Implementing support for APNS content-mutable field

hiranya911:
Implements support for `content-mutable` option in APNS messages.

```
type Aps struct {
  ContentMutable bool
  CustomData    map[string]interface{}
}
```


messaging/messaging.go
```diff
@@ -214,36 +214,52 @@ func (p *APNSPayload) MarshalJSON() ([]byte, error) {
...
-	type apsAlias Aps
-	s := &struct {
-		Alert            interface{} `json:"alert,omitempty"`
-		ContentAvailable *int        `json:"content-available,omitempty"`
-		*apsAlias
-	}{
-		apsAlias: (*apsAlias)(a),
-	}
-
+func (a *Aps) standardFields() map[string]interface{} {
```
bklimt: Would be nice to add a short comment for this function and mention that it doesn't include CustomData.

PR #127, opened 2018-03-29
Reading FCM error code from details section

hiranya911:
Porting the error handling improvement over from https://github.com/firebase/firebase-admin-node/pull/246

### DIFF
--- a/messaging/messaging.go
+++ b/messaging/messaging.go
@@ -381,7 +381,11 @@ type fcmResponse struct {
 
 type fcmError struct {
 	Error struct {
-		Status string `json:"status"`
+		Status  string `json:"status"`
+		Details []struct {
+			Type      string `json:"@type"`
+			ErrorCode string `json:"errorCode"`
+		}
 	} `json:"error"`
 }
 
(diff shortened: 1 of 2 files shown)


nordicdyno:
Any plans to expose Error.Details to client code?  I badly need it.

hiranya911:
No plans to expose `Error.Details`. Can you explain your use case a bit more in detail? Error handling in the Go SDK is about to undergo some changes, and we can consider new requirements when we get to that.

nordicdyno:
I want have ability to know which kind of error is happened and process it accordingly. Simple example:  group errors by type in monitoring counters.

Also it 's good Go practice to use typed errors not just simple error strings (that's ok for simple programs, but not for public APIs).

* https://blog.golang.org/error-handling-and-go

hiranya911:
That's the direction we are heading. But I can't comment on the timeline yet.

nordicdyno:
Nice! I'm glad to hear that. Right now I can patch library code by myself, but it would be much better to avoid it when API improved.

PR #131, opened 2018-04-10
Implementing Functions to Test Error Codes

hiranya911:
This PR improves error handling in the SDK by providing explicit functions that test for various error codes:

```
firebase.google.com/go/messaging
  func IsInternal(err error) bool
  func IsInvalidAPNSCredentials(err error) bool
  func IsInvalidArgument(err error) bool
  func IsMessageRateExceeded(err error) bool
  func IsMismatchedCredential(err error) bool
  func IsRegistrationTokenNotRegistered(err error) bool
  func IsServerUnavailable(err error) bool
  func IsTooManyTopics(err error) bool
  func IsUnknown(err error) bool
firebase.google.com/go/auth
  func IsEmailAlreadyExists(err error) bool
  func IsIDTokenRevoked(err error) bool
  func IsInsufficientPermission(err error) bool
  func IsPhoneNumberAlreadyExists(err error) bool
  func IsProjectNotFound(err error) bool
  func IsUIDAlreadyExists(err error) bool
  func IsUnknown(err error) bool
  func IsUserNotFound(err error) bool
```

### DIFF
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -436,12 +436,13 @@ var serverError = map[string]string{
 	"PROJECT_NOT_FOUND":       projectNotFound,
 }
 
-func handleServerError(err error) *internal.FirebaseError {
-	var serverCode string
+func handleServerError(err error) error {
 	gerr, ok := err.(*googleapi.Error)
-	if ok {
-		serverCode = gerr.Message
+	if !ok {
+		// Not a back-end error
+		return err
 	}
+	serverCode := gerr.Message
 	clientCode, ok := serverError[serverCode]
 	if !ok {
 		clientCode = unknown
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -23,8 +23,10 @@ import (
 	"strings"
 	"time"
 
+	"firebase.google.com/go/internal"
 	"golang.org/x/net/context"
 
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/identitytoolkit/v3"
 	"google.golang.org/api/iterator"
 )
@@ -215,8 +217,10 @@ func (c *Client) DeleteUser(ctx context.Context, uid string) error {
 
 	call := c.is.Relyingparty.DeleteAccount(request)
 	c.setHeader(call)
-	_, err := call.Context(ctx).Do()
-	return err
+	if _, err := call.Context(ctx).Do(); err != nil {
+		return handleServerError(err)
+	}
+	return nil
 }
 
 // GetUser gets the user data corresponding to the specified user ID.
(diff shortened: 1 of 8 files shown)


auth/user_mgt.go
```diff
@@ -345,10 +349,7 @@ func processClaims(p map[string]interface{}) error {
 		return nil
 	}
 
-	claims, ok := cc.(map[string]interface{})
-	if !ok {
```
bklimt: Are you sure you don't want to leave some kind of error reported here?

auth/user_mgt.go
```diff
@@ -372,6 +373,82 @@ func processClaims(p map[string]interface{}) error {
...
+	"CONFIGURATION_NOT_FOUND": projectNotFound,
+	"DUPLICATE_EMAIL":         emailAlredyExists,
+	"DUPLICATE_LOCAL_ID":      uidAlreadyExists,
+	"EMAIL_EXISTS":            emailAlredyExists,
+	"INSUFFICIENT_PERMISSION": insufficientPermission,
+	"PHONE_NUMBER_EXISTS":     phoneNumberAlreadyExists,
+	"PROJECT_NOT_FOUND":       projectNotFound,
+}
+
+func handleServerError(err error) *internal.FirebaseError {
```
bklimt: I would probably use `error` for the return type here, just because I've been bitten too many times by trying to use a specific error type.

auth/user_mgt.go
```diff
@@ -372,6 +373,82 @@ func processClaims(p map[string]interface{}) error {
...
+	var serverCode string
+	gerr, ok := err.(*googleapi.Error)
+	if ok {
+		serverCode = gerr.Message
+	}
+	clientCode, ok := serverError[serverCode]
+	if !ok {
+		clientCode = unknown
+	}
+	return internal.Error(clientCode, err.Error())
```
bklimt: It seems like this might get called in cases where the error is generated on the client by a failure to connect or something. Do we want to always wrap error in our own type like this, possibly losing original error data other than the string? Or should we just pass on errors that aren't from the backend as-is? This is something I've been thinking about a lot on all of our SDKs.

internal/internal.go
```diff
@@ -59,18 +61,59 @@ type StorageConfig struct {
...
+func Errorf(code string, msg string, args ...interface{}) *FirebaseError {
+	return Error(code, fmt.Sprintf(msg, args...))
+}
+
+// Code extracts the error code from the given error.
+//
+// If the error does not contain an error code, this function returns an empty string. Use this
+// internal function for unit tests in packages that cannot directly import the root firebase
+// package.
+func Code(err error) string {
```
bklimt: It looks like this isn't actually used anywhere?

auth/user_mgt.go
hiranya911: This is guaranteed to be a map since we have typed functions for setting custom claims (`SetCustomClaims()` for instance).

auth/user_mgt.go
hiranya911: +1. Done

internal/internal.go
hiranya911: Removed

auth/user_mgt.go
hiranya911: Good catch. We shouldn't be wrapping non-backend errors. Updated the implementation.

PR #134, opened 2018-04-17
Cleaning up the auth package

hiranya911:
Some minor syntactic changes are being made to the v3.0 branch in anticipation of the upcoming cookie management API.

* Renamed and moved some of the internal functions/variables around for clarity.
* Updated some of the error messages to be more Go-idiomatic

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -59,10 +59,12 @@ type Token struct {
 }
 
 func (t *Token) decodeFrom(s string) error {
+	// Decode into a regular map to access custom claims.
 	claims := make(map[string]interface{})
 	if err := decode(s, &claims); err != nil {
 		return err
 	}
+	// Now decode into Token to access the standard claims.
 	if err := decode(s, t); err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


auth/auth.go
```diff
@@ -58,6 +58,22 @@ type Token struct {
 	Claims   map[string]interface{} `json:"-"`
 }
 
+func (t *Token) decodeFrom(s string) error {
+	claims := make(map[string]interface{})
+	if err := decode(s, &claims); err != nil {
+		return err
+	}
+	if err := decode(s, t); err != nil {
```
bklimt: I think it's worth adding a comment explaining why you need to decode twice.

auth/auth.go
```diff
@@ -210,36 +227,36 @@ func (c *Client) VerifyIDToken(ctx context.Context, idToken string) (*Token, err
 		return nil, err
 	}
 
-	projectIDMsg := "Make sure the ID token comes from the same Firebase project as the credential used to" +
+	projectIDMsg := "make sure the id token comes from the same Firebase project as the credential used to" +
 		" authenticate this SDK."
-	verifyTokenMsg := "See https://firebase.google.com/docs/auth/admin/verify-id-tokens for details on how to " +
-		"retrieve a valid ID token."
+	verifyTokenMsg := "see https://firebase.google.com/docs/auth/admin/verify-id-tokens for details on how to " +
+		"retrieve a valid id token."
```
bklimt: Technically, go error messages shouldn't end with punctuation.

auth/auth.go
```diff
@@ -210,36 +227,36 @@ func (c *Client) VerifyIDToken(ctx context.Context, idToken string) (*Token, err
 		return nil, err
 	}
 
-	projectIDMsg := "Make sure the ID token comes from the same Firebase project as the credential used to" +
+	projectIDMsg := "make sure the id token comes from the same Firebase project as the credential used to" +
 		" authenticate this SDK."
```
bklimt: Technically, go error messages shouldn't end with punctuation.

auth/auth.go
hiranya911: Done

auth/auth.go
hiranya911: Done

auth/auth.go
hiranya911: Done

auth/auth.go
```diff
@@ -58,6 +58,24 @@ type Token struct {
...
+	claims := make(map[string]interface{})
+	if err := decode(s, &claims); err != nil {
+		return err
+	}
+	// Now decode into Token to access the standard claims.
+	if err := decode(s, t); err != nil {
+		return err
+	}
+
+	for _, r := range []string{"iss", "aud", "exp", "iat", "sub", "uid"} {
```
bklimt: Maybe another comment here like "Delete the standard claims from the custom claims map."

hiranya911: Done

PR #136, opened 2018-04-21
Cleaning up the auth (user management) code

hiranya911:
Since we migrated the user management API to use the Google API client (`identitytoolkit` package), this code has become bit of a readability nightmare:

* Unnecessary indirection: We store all user parameters in maps and then "apply" those values to `identitytoolkit` types (see `UserToCreate` and `UserToUpdate` types).
* Too many small/redundant helper functions.
* Heavy use of reflection.

This PR addresses the above issues by:

* Directly applying user parameters to the corresponding `identitytoolkit` types. Intermediate maps are no longer used.
* Removing most of the redundant helpers, and performing validations explicitly where needed.
* Removing all usages of reflection (these were mostly present in the helper functions).

### DIFF
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -97,7 +97,7 @@ type UserIterator struct {
 
 // UserToCreate is the parameter struct for the CreateUser function.
 type UserToCreate struct {
-	_req        *identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest
+	createReq   *identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest
 	uid         bool
 	displayName bool
 	email       bool
@@ -106,10 +106,10 @@ type UserToCreate struct {
 }
 
 func (u *UserToCreate) request() *identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest {
-	if u._req == nil {
-		u._req = &identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest{}
+	if u.createReq == nil {
+		u.createReq = &identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest{}
 	}
-	return u._req
+	return u.createReq
 }
 
 func (u *UserToCreate) validatedRequest() (*identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest, error) {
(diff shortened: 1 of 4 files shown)


auth/user_mgt.go
```diff
@@ -107,83 +97,236 @@ type UserIterator struct {
...
+func (u *UserToCreate) request() *identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest {
+	if u._req == nil {
+		u._req = &identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest{}
 	}
-	u.params[key] = value
+	return u._req
+}
+
+func (u *UserToCreate) validatedRequest() (*identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest, error) {
+	req := u.request() // it is allowed to create a user without any parameters
```
bklimt: Tiny nit:
// creating a user without any parameters is allowed

Otherwise, it's ambiguous what "it" means.

auth/user_mgt.go
```diff
@@ -107,83 +97,236 @@ type UserIterator struct {
 
 // UserToCreate is the parameter struct for the CreateUser function.
 type UserToCreate struct {
-	params map[string]interface{}
+	_req   *identitytoolkit.IdentitytoolkitRelyingpartySignupNewUserRequest
```
bklimt: I'm not familiar with prefixing names with underscore in go. Isn't that a python convention?

auth/user_mgt.go
```diff
@@ -107,83 +97,236 @@ type UserIterator struct {
...
+	fields struct {
```
bklimt: Is there a reason these are in a custom struct instead of just at the top level of the UserToCreate struct? If so, maybe a comment here explaining why would be helpful.

auth/user_mgt.go
```diff
@@ -452,22 +571,21 @@ func handleServerError(err error) error {
...
+	if len(val) == 0 {
 		return fmt.Errorf("photo url must be a non-empty string")
 	}
 	return nil
 }
 
-func validateEmail(val interface{}) error {
-	email := val.(string)
+func validateEmail(email string) error {
 	if email == "" {
```
bklimt: It's just slightly strange that you use `len(s) == 0` above but use `s == ""` here. I'm not sure which one would be preferred, but it would be nice to be consistent. Also, the actual error string is worded slightly differently.

auth/user_mgt.go

auth/user_mgt.go
hiranya911: I suppose the level of indirection can be removed without taking a hit on the readability. Removed.

auth/user_mgt.go
hiranya911: Changed to `== ""`

auth/user_mgt.go
hiranya911: I wanted to somehow indicate in the code that this field should not be accessed directly. It should always be accessed via `request()` or `validatedRequest()` functions. This also works well with IDE assist, since typing `u.req` auto completes to `u.request()`. WDYT?

auth/user_mgt.go
bklimt: It's just slightly odd that "photo url must be a non-empty string", but "email must not be empty". Minor nit, though.

Issue #138, opened 2018-04-29
FR: Support getting multiple user records at once by email


CDDelta:
### [READ] Step 1: Are you in the right place?

  * For issues or feature requests related to __the code in this repository__
    file a GitHub issue.
    * If this is a __feature request__ make sure the issue title starts with "FR:".
  * For general technical questions, post a question on [StackOverflow](http://stackoverflow.com/)
    with the firebase tag.
  * For general Firebase discussion, use the [firebase-talk](https://groups.google.com/forum/#!forum/firebase-talk)
    google group.
  * For help troubleshooting your application that does not fall under one
    of the above categories, reach out to the personalized
    [Firebase support channel](https://firebase.google.com/support/).

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: -
  * Firebase SDK version: -
  * Library version: 2.7.0
  * Firebase Product: auth (auth, database, storage, etc)

### [REQUIRED] Step 3: Describe the problem
Firebase auth admin SDK does not support getting multiple user records at once.

#### Steps to reproduce:
-

What happened? How can we make the problem occur?
This could be a description, log/console output, etc.

#### Relevant Code:

```
// TODO(you): code here to reproduce the problem
```

vr001:
Submitted a PR https://github.com/firebase/firebase-admin-go/pull/140. 

cc @CDDelta

hiranya911:
@vr001 can you describe your exact use case a little bit? In what circumstances would you need this feature? Knowing this will allow us to convince the rest of the Firebase team that this is something worth pursuing.

Also, is the same required for finding users by uid and phone number?

vr001:
My use case is almost similar to what CDDelta mentioned. @hiranya911

hiranya911:
Auth team is doing some work to enable this feature. Stay tuned.

CDDelta:
Hi @hiranya911, do you have any updates to share regarding this feature? Thanks!

PR #140, opened 2018-05-06
adding GetUsersByEmail feature

vr001:

### Discussion

With reference to issue https://github.com/firebase/firebase-admin-go/issues/138. 

### Testing

Unit test and integration tests passed, but haven't added test for the new function GetUsersByEmail.

### DIFF
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -256,6 +256,25 @@ func (c *Client) GetUserByEmail(ctx context.Context, email string) (*UserRecord,
 	return c.getUser(ctx, request)
 }
 
+// GetUsersByEmail gets the users data corresponding to the specified emails.
+func (c *Client) GetUsersByEmail(ctx context.Context, emails []string) ([]*UserRecord, error) {
+	var userRecords []*UserRecord
+	for _, email := range emails {
+		if err := validateEmail(email); err != nil {
+			return nil, err
+		}
+		request := &identitytoolkit.IdentitytoolkitRelyingpartyGetAccountInfoRequest{
+			Email: []string{email},
+		}
+		UserRec, err := c.getUser(ctx, request)
+		if err != nil {
+			return nil, err
+		}
+		userRecords = append(userRecords, UserRec)
+	}
+	return userRecords, nil
+}
+
 // Users returns an iterator over Users.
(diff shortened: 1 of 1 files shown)


CDDelta:
Thanks for doing this but I think there are some ways to optimize this function.

auth/user_mgt.go
```diff
@@ -256,6 +256,25 @@ func (c *Client) GetUserByEmail(ctx context.Context, email string) (*UserRecord,
...
 
+// GetUsersByEmail gets the users data corresponding to the specified emails.
+func (c *Client) GetUsersByEmail(ctx context.Context, emails []string) ([]*UserRecord, error) {
+	var userRecords []*UserRecord
+	for _, email := range emails {
+		if err := validateEmail(email); err != nil {
+			return nil, err
+		}
+		request := &identitytoolkit.IdentitytoolkitRelyingpartyGetAccountInfoRequest{
+			Email: []string{email},
```
CDDelta: You can pass the entire slice of emails to the request and get all the account info back at once.

vr001: I am using the existing function https://github.com/firebase/firebase-admin-go/blob/master/auth/user_mgt.go#L642 whose return type is *UserRecord. 
As per what you suggested, probably I would need to create a separate function which returns []*UserRecord. Let me know your thoughts. Thanks.

CDDelta: Yes, I think you would have to do that. I am unsure how you would have to format the errors though. Maybe someone from the Firebase team can chime in here?

hiranya911:
Thanks @vr001  for the PR. A couple of points to note:

1. Since this is a new API, it has to go through our internal review processes. This will take some time, so please sit tight.
2. Like @CDDelta point out, we should try to fetch all the necessary data from just one RPC call. The way it is implemented now is not going to scale.

vr001:
Thanks @hiranya911 for your response. 
I agree with the one RPC call for fetching all the users. Looking forward to hear regarding the plan (or proposal) for the new API.

hiranya911:
The Auth team has a plan to implement this feature. That will enable querying multiple user IDs, emails and phone numbers in a single request. Closing this.

PR #146, opened 2018-05-09
Import Users API for auth.Client

hiranya911:
Introduces the following new API:

```
ImportUsers(ctx context.Context, users []*auth.UserToImport, opts ...auth.UserImportOption) (*auth.UserImportResult, error)
```

Allows importing users into Firebase Auth in bulk (1000 users per call).

### DIFF
--- a/auth/hash/hash.go
+++ b/auth/hash/hash.go
@@ -50,3 +50,23 @@ func (s *Scrypt) Config() (*internal.HashConfig, error) {
 		MemoryCost:    int64(s.MemoryCost),
 	}, nil
 }
+
+// HMACSHA512 represents the HMAC SHA512 hash algorithm.
+type HMACSHA512 struct {
+	Key []byte
+}
+
+// Config returns the validated hash configuration.
+func (h *HMACSHA512) Config() (*internal.HashConfig, error) {
+	return hmacConfig("HMAC_SHA512", h.Key)
+}
+
+func hmacConfig(name string, key []byte) (*internal.HashConfig, error) {
+	if len(key) == 0 {
+		return nil, errors.New("signer key not specified")
+	}
+	return &internal.HashConfig{
+		HashAlgorithm: name,
+		SignerKey:     base64.RawURLEncoding.EncodeToString(key),
+	}, nil
+}
--- a/auth/hash/hash.go
+++ b/auth/hash/hash.go
@@ -1,0 +1,34 @@
+package hash // import "firebase.google.com/go/auth/hash"
+
+import (
+	"encoding/base64"
+	"errors"
+
+	"firebase.google.com/go/internal"
+)
+
+type Scrypt struct {
+	Key           []byte
+	SaltSeparator []byte
+	Rounds        int
+	MemoryCost    int
+}
+
+func (s *Scrypt) Config() (*internal.HashConfig, error) {
+	if len(s.Key) == 0 {
+		return nil, errors.New("signer key not specified")
+	}
+	if s.Rounds < 1 || s.Rounds > 8 {
+		return nil, errors.New("rounds must be between 1 and 8")
+	}
+	if s.MemoryCost < 1 || s.MemoryCost > 14 {
+		return nil, errors.New("memory cost must be between 1 and 14")
+	}
+	return &internal.HashConfig{
+		HashAlgorithm: "SCRYPT",
+		SignerKey:     base64.RawURLEncoding.EncodeToString(s.Key),
+		SaltSeparator: base64.RawURLEncoding.EncodeToString(s.SaltSeparator),
+		Rounds:        int64(s.Rounds),
+		MemoryCost:    int64(s.MemoryCost),
+	}, nil
+}
--- a/auth/hash/hash.go
+++ b/auth/hash/hash.go
@@ -13,7 +13,8 @@
 // limitations under the License.
 
 // Package hash contains a collection of password hash algorithms that can be used with the
-// auth.ImportUsers() API.
+// auth.ImportUsers() API. Refer to https://firebase.google.com/docs/auth/admin/import-users for
+// more details about supported hash algorithms.
 package hash // import "firebase.google.com/go/auth/hash"
 
 import (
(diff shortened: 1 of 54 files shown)


auth/hash/hash.go
```diff
@@ -0,0 +1,203 @@
...
+type Bcrypt struct{}
+
+// Config returns the validated hash configuration.
+func (b *Bcrypt) Config() (*internal.HashConfig, error) {
+	return &internal.HashConfig{HashAlgorithm: "BCRYPT"}, nil
+}
+
+// StandardScrypt represents the standard scrypt hash algorithm.
+type StandardScrypt struct {
+	BlockSize        int
```
bklimt: It seems like the zero values for these params aren't useful. Do we wanna provide any guidance on reasonable defaults?

auth/hash/hash.go
```diff
@@ -0,0 +1,203 @@
...
+type MD5 struct {
+	Rounds int
+}
+
+// Config returns the validated hash configuration.
+func (h *MD5) Config() (*internal.HashConfig, error) {
+	return basicConfig("MD5", h.Rounds)
+}
+
+// PBKDF2SHA256 represents the PBKDF2SHA256 hash algorithm.
```
bklimt: These comments feel unhelpful. Is there anything more reasonable we could put? Like maybe a link to more info about choosing a hash algorithm?

auth/hash/hash.go
```diff
@@ -0,0 +1,203 @@
...
+	"errors"
+
+	"firebase.google.com/go/internal"
+)
+
+// Bcrypt represents the BCRYPT hash algorithm.
+type Bcrypt struct{}
+
+// Config returns the validated hash configuration.
+func (b *Bcrypt) Config() (*internal.HashConfig, error) {
```
bklimt: These structs are all pretty small, and it makes me wonder if we want (b *Bcrypt) or just (b Bcrypt). Are we basing this API design on any particular precedent?

auth/user_mgt.go
```diff
@@ -84,6 +86,217 @@ type ExportedUserRecord struct {
...
+		}
+		if p.ProviderId == "" {
+			return nil, fmt.Errorf("user provider must specify a provider ID")
+		}
+	}
+	return info, nil
+}
+
+// UID setter. This field is required.
+func (u *UserToImport) UID(uid string) *UserToImport {
```
bklimt: Setters seem like an anti-pattern in go. It seems like you're making a fluent API, but then it's a little weird for a fluent API to mutate. Why are we providing this API rather than letting them supply a struct literal that's filled in already?

auth/user_mgt.go
```diff
@@ -84,6 +86,217 @@ type ExportedUserRecord struct {
...
+//
+// The Index field corresponds to the index of the failed user in the users array that was passed
+// to ImportUsers().
+type ErrorInfo struct {
+	Index  int
+	Reason string
+}
+
+// UserImportResult represents the result of an ImportUsers() call.
+type UserImportResult struct {
```
bklimt: How does this interface compare to other Firebase APIs for batch operations? Does Firestore's batch API return a similar struct?

auth/user_mgt.go
```diff
@@ -475,6 +688,59 @@ func (c *Client) SetCustomUserClaims(ctx context.Context, uid string, customClai
 	return c.updateUser(ctx, uid, (&UserToUpdate{}).CustomClaims(customClaims))
 }
 
+// ImportUsers imports an array of users to Firebase Auth.
+//
+// No more than 1000 users can be imported in a single call. If at least one user specifies a
+// password, a UserImportHash must be specified as an option.
+func (c *Client) ImportUsers(ctx context.Context, users []*UserToImport, opts ...UserImportOption) (*UserImportResult, error) {
+	if len(users) == 0 {
+		return nil, fmt.Errorf("users list must not be empty")
```
bklimt: I'd recommend using `errors.New()` instead of `fmt.Errorf()` when you don't need any interpolation.

auth/user_mgt.go
```diff
@@ -475,6 +688,59 @@ func (c *Client) SetCustomUserClaims(ctx context.Context, uid string, customClai
...
+		req.Users = append(req.Users, vu)
+	}
+
+	for _, opt := range opts {
+		if err := opt.applyTo(req); err != nil {
+			return nil, err
+		}
+	}
+	if hashRequired && req.HashAlgorithm == "" {
+		return nil, fmt.Errorf("hash algorithm option is required to import users with passwords")
```
bklimt: `errors.New`

auth/hash/hash.go
hiranya911: Added links to the public documentation.

auth/hash/hash.go
hiranya911: Added links to the public documentation. It's still not perfect, but way better than just spelling out the type names. We can work with the auth team to get the documentation improved further.

auth/hash/hash.go
hiranya911: Switched to implementing the interface on the struct (non-pointer) types. This works well for small structs like the ones defined in this package. Also as we found out, developers can pass the values by value or by pointer in this case, and both would work.

auth/user_mgt.go
hiranya911: We've used this for `UserToCreate` and `UserToUpdate` before. This enables us to handle default values of fields well (e.g. `EmailVerified`). I think in the future we can provide an unmarshal method to easily construct a user from a json string.

auth/user_mgt.go
hiranya911: This is similar to https://godoc.org/firebase.google.com/go/messaging#TopicManagementResponse

Firestore seems to return a plain `WriteResult`: https://godoc.org/cloud.google.com/go/firestore#WriteBatch.Commit

auth/user_mgt.go

auth/user_mgt.go
bklimt: SGTM

auth/user_mgt.go
bklimt: Well, it's good to be consistent with UserToCreate and UserToUpdate, so SGTM.

PR #151, opened 2018-05-25
Improved error handling in iid package

### LINKED ISSUE #149: FR: return identifiable errors in iid package
InstanceID client returns a formatted error message: https://github.com/firebase/firebase-admin-go/blob/master/iid/iid.go#L88

Caller cannot distinguish the error and take action. Errors should be public and returned error should be the error itself.

```go
if err, ok := errorCodes[resp.Status]; ok {
	return err
}
```

hiranya911:
Resolves #149 

go/firebase-go-iid-errors

### DIFF
--- a/iid/iid.go
+++ b/iid/iid.go
@@ -29,15 +29,77 @@ import (
 
 const iidEndpoint = "https://console.firebase.google.com/v1"
 
-var errorCodes = map[int]string{
-	http.StatusBadRequest:          "malformed instance id argument",
-	http.StatusUnauthorized:        "request not authorized",
-	http.StatusForbidden:           "project does not match instance ID or the client does not have sufficient privileges",
-	http.StatusNotFound:            "failed to find the instance id",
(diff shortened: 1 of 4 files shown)


iid/iid.go
```diff
@@ -84,8 +146,11 @@ func (c *Client) DeleteInstanceID(ctx context.Context, iid string) error {
 		return err
 	}
 
-	if msg, ok := errorCodes[resp.Status]; ok {
-		return fmt.Errorf("instance id %q: %s", iid, msg)
+	if info, ok := errorCodes[resp.Status]; ok {
```
bklimt: I notice we're only using the http status code here. Can't our APIs return more specific error codes in the body of the response? Or am I thinking of something else?

hiranya911: The endpoint we use for this functionality today signals errors via HTTP status code. In the future the backend service will send more detailed and well-formed error responses.

PR #152, opened 2018-05-29
Refactoring the user management integration tests

### LINKED ISSUE #145: Rewrite user management integration tests using random UIDs
Rewrite user management integration tests using random UIDs instead of fixed ones. With the current approach, whenever a test fails, we need to manually clean up all the user accounts before we can launch the tests again.

hiranya911:
The existing test suite is essentially one big test case composed of many helper functions. It also keeps a lot of intermediate state (fixtures), and tends to leave the Firebase project in an inconsistent state on failures. This PR:

1. breaks the test suite into multiple top-level test cases
2. uses random user IDs in each test case to avoid keeping state, and prevent collisions between different runs
3. uses the `defer` keyword to ensure graceful tear down

Resolves #145

### DIFF
--- a/integration/auth/auth_test.go
+++ b/integration/auth/auth_test.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
-	"math/rand"
 	"net/http"
 	"os"
 	"testing"
@@ -48,23 +47,19 @@ func TestMain(m *testing.M) {
 		os.Exit(0)
 	}
 
-	ctx := context.Background()
-	app, err := internal.NewTestApp(ctx, nil)
+	app, err := internal.NewTestApp(context.Background(), nil)
 	if err != nil {
 		log.Fatalln(err)
 	}
-
-	client, err = app.Auth(ctx)
+	client, err = app.Auth(context.Background())
 	if err != nil {
 		log.Fatalln(err)
 	}
-	rand.Seed(time.Now().UnixNano())
 	os.Exit(m.Run())
 }
 
 func TestCustomToken(t *testing.T) {
 	ct, err := client.CustomToken(context.Background(), "user1")
-
 	if err != nil {
 		t.Fatal(err)
 	}
@@ -72,6 +67,7 @@ func TestCustomToken(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
+	defer deleteUser("user1")
 
 	vt, err := client.VerifyIDToken(context.Background(), idt)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


integration/auth/user_mgt_test.go
```diff
@@ -527,14 +396,45 @@ func testImportUsersWithPassword(t *testing.T) {
...
 }
 
-const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
-
 var seededRand = rand.New(rand.NewSource(time.Now().UnixNano()))
 
-func randomString(length int) string {
-	b := make([]byte, length)
+func randomUID() string {
+	var letters = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789")
```
bklimt: Can this be const?

integration/auth/user_mgt_test.go
```diff
@@ -527,14 +396,45 @@ func testImportUsersWithPassword(t *testing.T) {
...
+	b := make([]rune, 32)
 	for i := range b {
-		b[i] = charset[seededRand.Intn(len(charset))]
+		b[i] = letters[seededRand.Intn(len(letters))]
 	}
 	return string(b)
 }
+
+func randomPhoneNumber() string {
+	var letters = []rune("0123456789")
```
bklimt: I would maybe name this `digits`.

integration/auth/auth_test.go
```diff
@@ -78,97 +76,87 @@ func TestCustomToken(t *testing.T) {
 	if vt.UID != "user1" {
 		t.Errorf("UID = %q; want UID = %q", vt.UID, "user1")
 	}
-	if err = client.DeleteUser(context.Background(), "user1"); err != nil {
-		t.Error(err)
+}
+
+func TestCustomTokenWithClaims(t *testing.T) {
+	ct, err := client.CustomTokenWithClaims(context.Background(), "user2", map[string]interface{}{
```
bklimt: Do you want to use random UIDs on all these other tests?

integration/auth/user_mgt_test.go
hiranya911: Arrays cannot be `const` in Go.

integration/auth/user_mgt_test.go
hiranya911: Done

integration/auth/auth_test.go
hiranya911: Done

integration/auth/auth_test.go
```diff
@@ -235,3 +223,7 @@ func postRequest(url string, req []byte) ([]byte, error) {
 	}
 	return ioutil.ReadAll(resp.Body)
 }
+
+func deleteUser(uid string) {
+	client.DeleteUser(context.Background(), uid)
```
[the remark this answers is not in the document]
hiranya911: This is intended to be a best effort delete operation. I don't think we want to fail a test because the clean up operation failed. Added a comment stating this.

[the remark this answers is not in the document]
bklimt: I still think it would be good to print out a warning when it fails.

PR #153, opened 2018-05-30
Adding snippets for auth.ImportUsers() API

hiranya911:
Snippets to appear on https://firebase.google.com/docs/auth/admin/import-users

### DIFF
--- a/snippets/auth.go
+++ b/snippets/auth.go
@@ -15,12 +15,14 @@
 package snippets
 
 import (
+	"encoding/base64"
 	"log"
 
 	"golang.org/x/net/context"
 
 	firebase "firebase.google.com/go"
 	"firebase.google.com/go/auth"
+	"firebase.google.com/go/auth/hash"
 	"google.golang.org/api/iterator"
 )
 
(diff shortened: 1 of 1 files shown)


snippets/auth.go
```diff
@@ -379,3 +381,196 @@ func listUsers(ctx context.Context, client *auth.Client) {
...
+			DisplayName("John Doe").
+			Email("johndoe@gmail.com").
+			PhotoURL("http://www.example.com/12345678/photo.png").
+			EmailVerified(true).
+			PhoneNumber("+11234567890").
+			CustomClaims(map[string]interface{}{"admin": true}). // set this user as admin
+			ProviderData([]*auth.UserProvider{                   // user with Google provider
+				{
+					UID:         "google-uid",
+					Email:       "johndoe@gmail.com",
```
bklimt: We probably want this to be @example.com.

snippets/auth.go
```diff
@@ -379,3 +381,196 @@ func listUsers(ctx context.Context, client *auth.Client) {
...
+	// [END import_with_scrypt]
+}
+
+func importWithoutPassword(ctx context.Context, client *auth.Client) {
+	// [START import_without_password]
+	users := []*auth.UserToImport{
+		(&auth.UserToImport{}).
+			UID("some-uid").
+			DisplayName("John Doe").
+			Email("johndoe@gmail.com").
```
bklimt: And here.

snippets/auth.go
hiranya911: The Node.js sample already published by the Auth team uses the `gmail` domain. This is just being consistent with that.

https://firebase.google.com/docs/auth/admin/import-users#import_users_without_passwords

PR #155, opened 2018-06-01
Support for creating custom tokens without service account credentials

hiranya911:
Currently the SDK must be initialized with service account credentials in order to be able to call FirebaseAuth.createCustomToken(). With this PR, the SDK will attempt to sign custom tokens by calling the IAM service in the cloud when the service account credentials are not provided.

```
// Following will work in GCP managed runtimes where the Metadata service is present
app, err := firebase.NewApp(ctx, nil)
...
token, err := client.CustomToken(uid)
```

```
// Following will work anywhere
app, err := firebase.NewApp(ctx, &firebase.Config{
    ServiceAccount: "test-service-account@project-id.iam.gserviceaccount.com",
})
...
token, err := client.CustomToken(uid)
```

I'm also refactoring the `auth` package a little bit by:
* Moving all token creation utils to a new `token_generator.go` file
* Moving all token verification utils to a new `token_verifier.go` file
* Removing `crypto.go` and `jwt.go` files

go/firebase-admin-sign
go/firebase-admin-iam-sign


auth/auth.go
```diff
@@ -98,40 +78,45 @@ type signer interface {
...
+		err    error
 	)
-	if c.Creds != nil && len(c.Creds.JSON) > 0 {
-		var svcAcct struct {
-			ClientEmail string `json:"client_email"`
-			PrivateKey  string `json:"private_key"`
-		}
-		if err := json.Unmarshal(c.Creds.JSON, &svcAcct); err != nil {
+	// Initialize a signer by following the go/firebase-admin-sign protocol.
+	if conf.Creds != nil && len(conf.Creds.JSON) > 0 {
```
bklimt: The description of this PR says that a nil conf will be allowed, but this would crash in that case. I think this is probably the right behavior, and the description is wrong. But it would be good to sync them up.

auth/user_mgt.go
```diff
@@ -822,6 +834,7 @@ var serverError = map[string]string{
 	"DUPLICATE_LOCAL_ID":      uidAlreadyExists,
 	"EMAIL_EXISTS":            emailAlredyExists,
 	"INSUFFICIENT_PERMISSION": insufficientPermission,
+	"PERMISSION_DENIED":       insufficientPermission,
```
bklimt: Interesting.

integration/auth/auth_test.go
```diff
@@ -161,6 +175,23 @@ func TestRevokeRefreshTokens(t *testing.T) {
 	}
 }
 
+func verifyCustomToken(t *testing.T, ct, uid string) *auth.Token {
+	idt, err := signInWithCustomToken(ct)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer deleteUser(uid)
```
bklimt: I would add a comment for this function clarifying that, in addition to verifying the token, this function creates and deletes the user with the given UID.

You know, it might actually make more sense to get the uid out of `ct`, if possible. Otherwise, there's the small chance of passing in the wrong uid.

auth/auth_test.go
```diff
@@ -86,25 +85,61 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		log.Fatalln(err)
 	}
-	client.ks = ks
+	client.keySource = ks
 
 	testGetUserResponse, err = ioutil.ReadFile("../testdata/get_user.json")
```
bklimt: Would it make sense to break this one out, too?

auth/user_mgt.go
bklimt: What I meant by that is that it seems a little odd for two different strings to map to the same error type. Is that intentional?

auth/auth.go
hiranya911: The conf passed to `NewApp()` can be nil. `NewAuth()` is an internal API, and always get called with a non-nil conf.

auth/user_mgt.go
hiranya911: We currently don't have a public permission denied error code: https://firebase.google.com/docs/auth/admin/errors

This will have to do for the moment. I'll run it by the auth team and see if we can get another public error code.

integration/auth/auth_test.go
hiranya911: Added comment. 

I'd prefer to avoid the token decoding step necessary to extract the uid from the custom token here. If the wrong uid gets passed in, the test will fail since we compare them later.

auth/auth.go
bklimt: Assuming by `NewAuth()` you meant `auth.NewClient()`, then SGTM.

auth/auth_test.go
hiranya911: This gets used around 3 times in the test suite. We can avoid a few additional file system reads by doing it here.

auth/user_mgt.go
hiranya911: Auth team is ok with this. We define `insufficient-permission` error code as follows:

```
The credential used to initialize the Admin SDK has insufficient permission to access the requested Authentication resource. See Add Firebase to your app for documentation on how to generate a credential with appropriate permissions and use it to authenticate the Admin SDKs.
```

Seems to fit the bill in this case.

hiranya911:
As discussed offline, Auth team is ok with using the same insufficient-permission error code in this case. Sending back for another look.

PR #164, opened 2018-07-24
Multi DB API

hiranya911:
Added `app.DatabaseWithURL()` API.


firebase_test.go
```diff
@@ -246,6 +246,10 @@ func TestDatabase(t *testing.T) {
 	if c, err := app.Database(ctx); c == nil || err != nil {
 		t.Errorf("Database() = (%v, %v); want (db, nil)", c, err)
 	}
+	url := "https://other-mock-db.firebaseio.com"
+	if c, err := app.DatabaseWithURL(ctx, url); c == nil || err != nil {
```
bklimt: Ideally, we would test to make sure the database url of the created client is the one passed in and not the default.

PR #165, opened 2018-07-27
Added Additional WebpushNotification Fields

hiranya911:
FCM Webpush API now supports additional parameters as defined in https://developer.mozilla.org/en-US/docs/Web/API/notification/Notification

go/firebase-admin-webpush


messaging/messaging_utils.go
```diff
@@ -126,6 +131,23 @@ func validateApsAlert(alert *ApsAlert) error {
 	return nil
 }
 
+func validateWebpushConfig(webpush *WebpushConfig) error {
+	if webpush == nil || webpush.Notification == nil {
+		return nil
+	}
+	dir := webpush.Notification.Direction
+	if dir != "" && dir != "ltr" && dir != "rtl" {
+		return fmt.Errorf("direction must be 'ltr' or 'rtl'")
```
mmermerkaya: `"auto"` should be allowed too. Even though it's already the default, the browser doesn't complain if you manually set it.

hiranya911: Good catch. Done.

PR #167, opened 2018-08-09
Added 'Image' and 'ClickAction' field to WebpushNotification

spinache:
Sending image in notification is possible


hiranya911:
This is addressed in #165

PR #175, opened 2018-09-13
Move token signature verification after fields

### LINKED ISSUE #174: auth/Client.VerifyIDToken should verify (some) fields before checking signature
Field validation in auth/Client.VerifyIDToken(), like checking `Expires` and `IssuedAt`, should be done before the, potentially expensive, operation [verifyToken](https://github.com/firebase/firebase-admin-go/blob/10442ce72ae5afa134bdea11670094a04cfe5805/auth/auth.go#L223). Deep down `verifyToken` uses locking and issues a http request. Although this is not the biggest overhead, it would be nice if e.g. expired tokens where not dependent on sub-requests to Firebase API (an expired token is not valid no matter the signature)

palsivertsen:
Fixes #174 

Token signature validation can be an expensive operation.
For tokens with invalid fields it is not necessary to check the
signature and it is therfore moved to the bottom.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -220,9 +220,6 @@ func (c *Client) VerifyIDToken(ctx context.Context, idToken string) (*Token, err
 		return nil, fmt.Errorf("id token must be a non-empty string")
 	}
 
-	if err := verifyToken(ctx, idToken, c.keySource); err != nil {
-		return nil, err
-	}
 	segments := strings.Split(idToken, ".")
 
 	var (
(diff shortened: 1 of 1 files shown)


palsivertsen:
I signed it!

hiranya911:
Code changes LGTM. Thanks @palsivertsen for putting this together. I'm wondering why the CLA check hasn't passed yet. Can you review the following instructions, and see everything is in order:

> It's possible we don't have your GitHub username or you're using a different email address on your commit. Check your existing CLA data and verify that your email is set on your git commits.

palsivertsen:
There where some issues with the CLA. I updated the document and CLA is pending. Hopefully processed today.

palsivertsen:
@hiranya911 can you recheck the CLA?

Issue #182, opened 2018-10-22
Missing WebpushFcmOptions in WebpushConfig


mmermerkaya:
There is a new field in the FCM Send API that allows the user to set a link. It would be great if the Admin SDK supported this.

See [the reference docs](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#webpushconfig) for details.

hiranya911:
Fixed in #185

PR #185, opened 2018-10-23
add fcmOptions in WebpushConfig

chemidy:
new option in webpushconfig 
#182

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,6 @@
 language: go
 
 go:
-  - 1.6.x
-  - 1.7.x
-  - 1.8.x
   - 1.9.x
   - "1.10.x"
   - master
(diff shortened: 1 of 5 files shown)


chemidy:
Formatting has changed on [go 1.11](https://golang.org/doc/go1.11#gofmt) check the reason here : https://github.com/golang/go/commit/542ea5ad91367d2b40589942cc5757a4d5f43f97

So formating has changed 
-		"to": "/topics/test-topic",
+		"to":                  "/topics/test-topic",

this is why the build failed

chemidy:
Maybe we can just gofmt ( ./.travis.gofmt.sh) for go >= 1.11 due to incompatible changes

hiranya911:
We currently don't test on go 1.11 so I think we can ignore the gofmt issue for now.

.travis.yml
```diff
@@ -20,14 +20,14 @@ go_import_path: firebase.google.com/go
 
 before_install:
     # Golint requires Go 1.7 or later.
-    - if ! [[ "$TRAVIS_GO_VERSION" =~ ^1\.6\.([0-9]+|x)$ ]]; then go get github.com/golang/lint/golint; fi
+    - if ! [[ "$TRAVIS_GO_VERSION" =~ ^1\.6\.([0-9]+|x)$ || "$TRAVIS_GO_VERSION" =~ ^1\.7\.([0-9]+|x)$ || "$TRAVIS_GO_VERSION" =~ ^1\.8\.([0-9]+|x)$ ]]; then go get golang.org/x/lint/golint; fi
```
hiranya911: cloud.google.com/go has stopped supporting go 1.8 and earlier, which means we can stop supporting them too. Can you drop those targets from the Travis config, and update this file accordingly?

messaging/messaging.go
```diff
@@ -225,6 +224,7 @@ type WebpushConfig struct {
 	Headers      map[string]string    `json:"headers,omitempty"`
 	Data         map[string]string    `json:"data,omitempty"`
 	Notification *WebpushNotification `json:"notification,omitempty"`
+	FcmOptions   *WebpushFcmOptions   `json:"fcm_options,omitempty"`
```
hiranya911: FCM team tells me that this json key should be called `fcmOptions`. `fcm_options` is a documentation error, and they are working on it.

chemidy: OK :) fixed

.travis.yml
chemidy: OK, now that we have removed support 1.6. Maybe we can update context Import `golang.org/x/netcontext` -> `context` and other tricks for 1.6 in another PR.

hiranya911: Yes, lets make those changes in a separate PR.

PR #186, opened 2018-10-25
Add header for opting into correct URL decoding

rockwotj:
There is a long standing bug (~years) in the RTDB url decoding where we URL decode query parameters twice. Since we can't simply fix the issue without breaking users, we allow an opt-in upgrade/fix of the correct behavior via a header `X-Firebase-Decoding: 1`. We hope to make this the default (correct) behavior at some point in the near future. Which then this header will not be needed anymore.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -115,6 +115,7 @@ func (c *Client) send(
 	if c.authOverride != "" {
 		opts = append(opts, internal.WithQueryParam(authVarOverride, c.authOverride))
 	}
+	opts = append(opts, internal.WithHeader("X-Firebase-Decoding", "1"))
 	return c.hc.Do(ctx, &internal.Request{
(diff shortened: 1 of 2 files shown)


db/db.go
```diff
@@ -115,6 +115,7 @@ func (c *Client) send(
 	if c.authOverride != "" {
 		opts = append(opts, internal.WithQueryParam(authVarOverride, c.authOverride))
 	}
+	opts = append(opts, internal.WithHeader("X-Firebase-Decoding", "1"))
```
hiranya911: Can you also update the tests by adding something to https://github.com/firebase/firebase-admin-go/blob/master/db/db_test.go#L272?

And ideally add a test case to `integration/db/db_test.go` as well?

PR #120, opened 2018-03-15
Support Go 1.6

tbpg:
Hey there! So you want to contribute to a Firebase SDK? 
Before you file this pull request, please read these guidelines:

### Discussion

See #119 and #110.

### Testing

  * Make sure all existing tests in the repository pass after your change.
    * Done.
  * If you fixed a bug or added a feature, add a new test to cover your code.
    * N/A.

### API Changes

  * At this time we cannot accept changes that affect the public API.  If you'd like to help 
    us make Firebase APIs better, please propose your change in an issue so that we 
    can discuss it together.
    * N/A.

There are several small, mechanical changes in this PR:
* Import Context from `/x/`.
* Don't use subtests.
* Convert test types to float64 since Go 1.6 can't parse numbers in exponential form (e.g. `-7e+07`) into ints.
* Use `t.Fatal` in the messaging integration test instead of `log.Fatal`. `log.Fatal` doesn't include the line number and is harder to debug.
* Switch `request.WithContext` to `ctxhttp.Do`.

See the individual commits for more detail. Integration and unit tests pass on my machine - hopefully Travis agrees.

Let me know if you prefer one commit per PR and I'll rebase into a single commit.

PR #190, opened 2018-11-13
remove support go 1.6

chemidy:
cloud.google.com/go has stopped supporting go 1.8 and earlier. We can now use context.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -23,7 +23,7 @@ import (
 
 	"google.golang.org/api/option"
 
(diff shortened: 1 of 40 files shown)


chemidy:
Maybe we can rollback on subtests removed for 1.6 support in https://github.com/firebase/firebase-admin-go/pull/120

hiranya911:
Looks like a couple of errors were made when merging against the dev. Should be pretty straightforward to fix.

Issue #173, opened 2018-09-07
FR: Add subtitle for iOS push notifications


dobegor:
### Describe your environment
  * Messaging/Notifications

### Describe the problem
We cannot send push notifications with subtitle for iOS due to [ApsAlert](https://godoc.org/firebase.google.com/go/messaging#ApsAlert) struct lacking a `subtitle` field.

FCM API and the APNS itself do support this field on iOS 9+ (and it's safely ignored on older versions).

hiranya911:
Will implement this eventually. Please provide a pull request if you want to speed things up.

hiranya911:
This is implemented. Will be included in next release. Thanks @chmidy.

PR #191, opened 2018-11-14
FR #173 : add subtitle for ApsAlert

chemidy:
I Added subtitle  Based on [APS doc ](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification#2943365) 

#173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- [added] `messaging.ApsAlert` type now supports subtitle in its payload.
+
 # v3.4.0
 
 - [added] `firebase.App` now provides a new `DatabaseWithURL()` function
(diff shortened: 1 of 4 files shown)


hiranya911:
@chemidy can you also add a quick entry to the CHANGELOG file before I merge this?

broady:
nit, for future reference: in idiomatic Go, this should be called "APSAlert". Acronyms are uppercase.

hiranya911:
Thanks @broady. We considered that. It's not clear from the Apple documentation whether `aps` is an acronym. We are fairly certain it doesn't stand for anything. It's often just referred to as the "aps dictionary". I've only seen [one place](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/PayloadKeyReference.html) where it's spelled out as APS.

PR #198, opened 2018-11-21
format import

chemidy:
gofmt on import


hiranya911:
@chemidy just a ping on the question abov. Do we need to do something extra to get CI to catch this sort of issues>

chemidy:
You do not have to change anything.

I'm using vscode and by default the formatTool is not gofmt but goreturns : 

```
  // Pick 'gofmt', 'goimports', 'goreturns' or 'goformat' to run on format.
  "go.formatTool": "goreturns",
```

Check this [link]( http://goinbigdata.com/goimports-vs-gofmt/) for more informations

[goreturns is based on goimports](https://github.com/sqs/goreturns)

Tkbagz:
What do you mean

hiranya911:
@chemidy it sounds like we should change our CI script (`.travis.gofmt.sh`) to use `goimports` instead of `gofmt`. But we can do that in a separate PR.

PR #211, opened 2018-12-21
Add a a 5min tolerance to clock skew

### LINKED ISSUE #210: FR:  Token verification clock skew
* Operating System version: Win 10
  * Firebase SDK version: ? 
  * Library version: v3.5.0
  * Firebase Product: auth

#### Steps to reproduce:

When I attempt to validate a token that was just created, it fails due to the token being created in the future.  This is presumably due to my clock being a couple milliseconds behind the firebase servers.

#### Relevant Code:
My server code
```
                tokenStr := "something just posted from the browser"
		token, err = authClient.VerifyIDToken(context.Background(), tokenStr)
		if err != nil{
			log.Printf("invalid token: %s", err)
		}
```
Firebase `auth.go` code
```
          else if payload.IssuedAt > clk.Now().Unix() {
		err = fmt.Errorf("ID token issued at future timestamp: %d", payload.IssuedAt)
	} 
```

The firebase code does a strict clock comparison with no tolerance for time skew.

I've been able to mitigate this by wrapping the call in a function with a backoff retry, but it would be nice if there was a first class method for managing clock skew.

```

func checkTokenWithSkewBackoff(tokenStr string) (*auth.Token, error) {
	var token *auth.Token
	var err error
	for _, d := range []time.Duration{0 * time.Millisecond, 10 * time.Millisecond, 100 * time.Millisecond, 1000 * time.Millisecond} {
		time.Sleep(d)
		token, err = authClient.VerifyIDToken(context.Background(), tokenStr)
		if err == nil || ! strings.HasPrefix(err.Error(), "ID token issued at future timestamp") {
			return token, err
		}
	}
	return token, err
}
```

hiranya911:
ID token verification currently has no tolerance to clock skew. In other languages we have a 5 minute constant tolerance. See [Python implementation](https://github.com/googleapis/google-auth-library-python/blob/6e0781bb118a097b2398712901292d014a781278/google/auth/jwt.py#L176-L189) for reference.

This PR:
1. Adds a 5 minute tolerance to both `iat` and `exp` when verifying ID tokens
2. Uses a `mockClock` during tests for stable test results
3. Moves the package-level `clock` variable into the `auth.Client` for better encapsulation

Resolves #210

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -33,14 +33,13 @@ const (
 	idTokenCertURL   = "https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com"
 	issuerPrefix     = "https://securetoken.google.com/"
 	tokenExpSeconds  = 3600
+	clockSkewSeconds = 300
 )
 
 var reservedClaims = []string{
 	"acr", "amr", "at_hash", "aud", "auth_time", "azp", "cnf", "c_hash",
 	"exp", "firebase", "iat", "iss", "jti", "nbf", "nonce", "sub",
 }
-
-var clk clock = &systemClock{}
 
 // Token represents a decoded Firebase ID token.
 //
@@ -67,6 +66,7 @@ type Client struct {
 	projectID string
 	signer    cryptoSigner
 	version   string
+	clock     clock
 }
 
 type signer interface {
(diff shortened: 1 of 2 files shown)


dchenk:
Thanks for this!

setu1421:
Has it been released yet?

PR #221, opened 2019-01-28
Refactoring Unit Tests

hiranya911:
I'm planning some refactoring work (especially in the auth package). In preparation for that I wanted to clean up and refactor our unit tests first.

1. Using `internal.Clock` and `internal.MockAccessTokenSource` in all packages.
2. Removing the App Engine dependencies in the auth tests (this was never properly maintained or verified).
3. Updating auth unit tests to use more local variables, and less package variables for clarity and readability. Specifically, tests no longer use a single, shared `auth.Client` instance that is set up at the package level.

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -30,68 +30,44 @@ import (
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/option"
 	"google.golang.org/api/transport"
-	"google.golang.org/appengine"
-	"google.golang.org/appengine/aetest"
 )
 
 var (
 	client              *Client
-	ctx                 context.Context
 	testIDToken         string
 	testGetUserResponse []byte
+	optsWithServiceAcct = []option.ClientOption{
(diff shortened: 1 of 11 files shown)


auth/auth_test.go
```diff
@@ -30,166 +30,220 @@ import (
...
-
-	testIDToken = getIDToken(nil)
 	os.Exit(m.Run())
 }
 
-func TestNewClientServiceAccountSigner(t *testing.T) {
+func TestNewClientWithServiceAccountCredentials(t *testing.T) {
+	creds, err := transport.Creds(context.Background(), optsWithServiceAcct...)
+	if err != nil {
+		log.Fatalln(err)
```
bklimt: Maybe t.Fatal(err) would be preferable? I'm not sure that it makes any difference, though...

hiranya911: Good point. I think we should be using `t.Fatal()` inside test cases.

PR #222, opened 2019-02-11
Extracting the Token Verification Logic into a New Type

hiranya911:
We are planning to implement session cookie verification API soon. This requires the same code currently being used to verify ID tokens. In order to reuse this logic easily, I'm extracting it out and putting it in a new type named `tokenVerifier`.

### DIFF
--- a/auth/auth_test.go
+++ b/auth/auth_test.go
@@ -38,8 +38,6 @@ var (
 	testSigner          cryptoSigner
 	testIDTokenVerifier *tokenVerifier
 
-	testClock           = &internal.MockClock{Timestamp: time.Now()}
-	testProjectID       = "mock-project-id"
 	optsWithServiceAcct = []option.ClientOption{
 		option.WithCredentialsFile("../testdata/service_account.json"),
 	}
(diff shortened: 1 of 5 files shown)


auth/auth_test.go
```diff
@@ -87,8 +89,8 @@ func TestNewClientWithServiceAccountCredentials(t *testing.T) {
 	if _, ok := client.signer.(*serviceAccountSigner); !ok {
 		t.Errorf("NewClient().signer = %#v; want = serviceAccountSigner", client.signer)
 	}
-	if client.projectID != creds.ProjectID {
-		t.Errorf("NewClient().projectID = %q; want = %q", client.projectID, creds.ProjectID)
+	if err := checkIDTokenVerifier(client.idTokenVerifier, creds.ProjectID); err != nil {
+		t.Errorf("NewClient().idTokenVerifier = %v; want = nil", err)
```
bklimt: This message seems a little odd. Are you sure this is what you meant?

auth/token_verifier.go
```diff
@@ -34,8 +34,213 @@ import (
...
+	}
+
+	if err = decode(segments[1], &payload); err != nil {
+		return nil, err
+	}
+
+	issuer := tv.issuerPrefix + tv.projectID
+	if header.KeyID == "" {
+		if payload.Audience == firebaseAudience {
+			err = fmt.Errorf("expected %s but got a custom token", tv.articledShortName)
```
bklimt: I think all of this would be much more idiomatic if there were no `err` variable and it instead used `return` in these cases, and you got rid of all the `else`s. That'll also make it clearer / ensure that there's no logic happening in this function once the error condition has been determined.

https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

auth/auth_test.go
hiranya911: Changed to `t.Errorf("NewClient().idTokenVerifier: %v", err)`

auth/token_verifier.go
hiranya911: Done

Issue #220, opened 2019-01-28
Using HTTP Client for User Management


hiranya911:
In #60 we changed the user management code to use the auto-generated identity toolkit client instead of the HTTP client. This was a useful an interesting experiment that yielded some positive results in the short term. But I feel in the long run it has caused a couple of issues:

1. The auto-generated clients are not updated regularly, preventing us from launching new APIs at the pace we want.
2. We lose control over the transport layer, which makes it difficult to implement optimizations like retries in a single location.

Therefore, I'd like to consider going back to the HTTP client for user management. We can do this in steps:

1. Refactor the user mgt code and implement the currently missing APIs (e.g. session cookie management) on top of HTTP client without changing the existing APIs.
2. Incrementally migrate the existing user management APIs to also use the HTTP client.

@chemidy what do you think?

chemidy:
I think this is a useful, we can deliver new functions very quickly 👍

hiranya911:
We launched session cookie management API using this mechanism. I'll work on getting other APIs migrated as well.

canhlinh:
@hiranya911 I've a question. Does firebase support GRPC for sending notifications ?

hiranya911:
It probably does. @chong-shao should know more.

chong-shao:
For v1 send API, currently it does not support gRPC.

canhlinh:
@chong-shao Is there any API version are supporting gRPC ?

chong-shao:
Hi @canhlinh, as far as I know there is no version that supports gRPC for now.

canhlinh:
@chong-shao Thanks

Issue #188, opened 2018-10-30
FR: Manage Session Cookies


chemidy:
Add Firebase Auth server-side session cookie management : https://firebase.google.com/docs/auth/admin/manage-cookies

hiranya911:
This is blocked on https://github.com/googleapis/google-api-go-client/tree/master/identitytoolkit/v3 to get updated. @bojeil-google has there been any progress towards getting those API descriptions updated?

wejick:
Need this feature too
Not sure if you can pick this up anytime soon, then can someone explain how to get the session cookie?
Looked onto node sdk it's using `identitytoolkit.googleapis.com` but can't find reference for `:createSessionCookie` API.

fitzoh:
per @jadekler in googleapis/google-api-go-client#263 it's blocked until the firebase team updates the `identitytoolkit` API

fproulx-eoscanada:
Will that be fixed soon . ?

rcarver:
Any update here? I'd really like to use this feature.

bojeil-google:
We apologize for the delay. We are still working on exposing these new APIs in our discovery doc.

hiranya911:
This has been released.

PR #226, opened 2019-03-05
Implement SessionCookie() API

hiranya911:
This is an attempt to implement the `SessionCookie()` API in the Go Admin SDK. This has been implemented by directly calling the backend REST API, instead of going through the auto-generated identity toolkit client (see #220).

Gets us closer to resolving #188

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -44,6 +44,7 @@ type Client struct {
 	is *identitytoolkit.Service
 	userManagementClient
 	idTokenVerifier *tokenVerifier
+	cookieVerifier  *tokenVerifier
 	signer          cryptoSigner
 	version         string
 	clock           internal.Clock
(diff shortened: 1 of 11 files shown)


auth/auth.go
```diff
@@ -98,11 +98,19 @@ func NewClient(ctx context.Context, conf *internal.AuthConfig) (*Client, error)
 	if err != nil {
 		return nil, err
 	}
+
+	version := "Go/Admin/" + conf.Version
 	return &Client{
+		userManagementClient: userManagementClient{
+			baseURL:    idToolkitEndpoint,
+			projectID:  conf.ProjectID,
+			version:    version,
```
bklimt: It feels like a code smell to duplicate `version` in two places, although it might not be a bad idea. It's unclear without knowing more about userManagementClient.

hiranya911: In the future we will have `userManagementClient` implement all the user management APIs. In that world we won't need `auth.Client.version` at all. Will add a comment to make it clear that this is just an interim step.

auth/user_mgt.go
```diff
@@ -629,3 +630,95 @@ func (c *Client) getUser(ctx context.Context, request *identitytoolkit.Identityt
 	}
 	return eu.UserRecord, nil
 }
+
+const idToolkitEndpoint = "https://identitytoolkit.googleapis.com/v1/projects"
+
+type userManagementClient struct {
```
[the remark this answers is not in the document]

hiranya911:
@bklimt added some comments to address the points mentioned. Ready for another look.

PR #230, opened 2019-03-15
Implemented VerifySessionCookie() API

### LINKED ISSUE #188: FR: Manage Session Cookies

hiranya911:
Implements the following APIs:

* `VerifySessionCookie()`
* `VerifySessionCookieAndCheckRevoked()`
* IsSessionCookieRevoked()`

Resolves #188

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - [added] Implemented `auth.SessionCookie()` function for creating
   new Firebase session cookies given an ID token.
-- [added] Implementd `auth.VerifySessionCookie()` and
+- [added] Implemented `auth.VerifySessionCookie()` and
   `auth.VerifySessionCookieAndCheckRevoked()` functions for verifying
   Firebase session cookies.
 - [added] Implemented HTTP retries for the `db` package. This package
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - [added] Implemented `auth.SessionCookie()` function for creating
   new Firebase session cookies given an ID token.
+- [added] Implementd `auth.VerifySessionCookie()` and
+  `auth.VerifySessionCookieAndCheckRevoked()` functions for verifying
+  Firebase session cookies.
 - [added] Implemented HTTP retries for the `db` package. This package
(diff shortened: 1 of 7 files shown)


hiranya911:
This is a follow up to #226

auth/auth.go
```diff
@@ -224,13 +231,61 @@ func (c *Client) VerifyIDTokenAndCheckRevoked(ctx context.Context, idToken strin
...
+	return p, nil
+}
+
+func (c *Client) checkRevoked(ctx context.Context, token *Token) (bool, error) {
+	user, err := c.GetUser(ctx, token.UID)
+	if err != nil {
+		return false, err
+	}
+
+	return token.IssuedAt*1000 < user.TokensValidAfterMillis, nil
```
bklimt: Does this line need to handle skew? Could use a comment either way.

integration/auth/auth_test.go
```diff
@@ -57,6 +58,7 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		log.Fatalln(err)
 	}
+	rand.Seed(time.Now().UTC().UnixNano())
```
bklimt: In tests that use random, it's a good practice to log the seed passed to the randomizer, so that if the test turns out to be flaky, you can replay the failed test run exactly by reusing the seed.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 - [added] Implemented `auth.SessionCookie()` function for creating
   new Firebase session cookies given an ID token.
+- [added] Implementd `auth.VerifySessionCookie()` and
```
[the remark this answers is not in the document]
hiranya911: Done

auth/auth.go
hiranya911: I feel like we should, but our other implementations of this API currently doesn't do it. I will leave it as is for now, but run it by the Auth team for some comments.

integration/auth/auth_test.go
hiranya911: Good idea. Done.

PR #235, opened 2019-03-29
Implemented GetUser* APIs with internal.HTTPClient

hiranya911:
As a continuation of #220, I'm moving the `GetUser()`, `GetUserByEmail()` and `GetUserByPhoneNumber()` API implementations off the `identitytoolkit` package, and on to our homegrown `internal.HTTPClient`.

### DIFF
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -46,15 +46,15 @@ func (c *Client) setHeader(ic identitytoolkitCall) {
 
 // UserInfo is a collection of standard profile information for a user.
 type UserInfo struct {
-	DisplayName string
-	Email       string
-	PhoneNumber string
-	PhotoURL    string
+	DisplayName string `json:"displayName,omitempty"`
+	Email       string `json:"email,omitempty"`
+	PhoneNumber string `json:"phoneNumber,omitempty"`
+	PhotoURL    string `json:"photoUrl,omitempty"`
 	// In the ProviderUserInfo[] ProviderID can be a short domain name (e.g. google.com),
 	// or the identity of an OpenID identity provider.
 	// In UserRecord.UserInfo it will return the constant string "firebase".
-	ProviderID string
-	UID        string
+	ProviderID string `json:"providerId,omitempty"`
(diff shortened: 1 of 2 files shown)


auth/user_mgt.go
```diff
@@ -652,6 +593,156 @@ type userManagementClient struct {
...
+	}
+	return c.getUser(ctx, phoneNumberQuery(phone))
+}
+
+type userQuery interface {
+	name() string
+	build() map[string]interface{}
+}
+
+type uidQuery string
```
bklimt: This isn't bad, but every one of these implementations is the same, except for a hard-coded string. What about this?

```
type userQuery {
	field string
	value string
}

func (q userQuery) description() string {
	return fmt.Sprintf("%s: %q", q.field, q.value)
}

func (q userQuery) build() map[string]interface{} {
	return map[string]interface{}{
		q.field: []string{string(q.value)},
	}
}
```

Then instead of like `phoneNumberQuery(phone)`, it's `userQuery{"phoneNumber", phone}`. The potential error message when it fails isn't quite as pretty, but it's a lot less code. That is, assuming there won't be other types of user queries.

I'm okay either way, but thought I'd suggest it.

hiranya911: I like the simplicity of this approach, so I'm implementing it with a minor modification to retain the nicer error messages. We are unlikely to introduce more complex queries in the future. But if we do, we can implement the interface-based solution at that time.

auth/user_mgt.go
```diff
@@ -652,6 +593,156 @@ type userManagementClient struct {
...
+// GetUserByPhoneNumber gets the user data corresponding to the specified user phone number.
+func (c *userManagementClient) GetUserByPhoneNumber(ctx context.Context, phone string) (*UserRecord, error) {
+	if err := validatePhone(phone); err != nil {
+		return nil, err
+	}
+	return c.getUser(ctx, phoneNumberQuery(phone))
+}
+
+type userQuery interface {
+	name() string
```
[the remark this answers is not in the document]

hiranya911:
@bklimt I've made the suggested change. PTAL.

PR #236, opened 2019-03-29
Added X-Firebase-Client header to FCM API calls

hiranya911:


messaging/messaging.go
```diff
@@ -807,7 +807,10 @@ func (c *Client) makeSendRequest(ctx context.Context, req *fcmRequest) (string,
 		Method: http.MethodPost,
 		URL:    fmt.Sprintf("%s/projects/%s/messages:send", c.fcmEndpoint, c.project),
 		Body:   internal.NewJSONEntity(req),
-		Opts:   []internal.HTTPOption{internal.WithHeader("X-GOOG-API-FORMAT-VERSION", "2")},
+		Opts: []internal.HTTPOption{
```
chong-shao: I'm not familiar about the go code style, maybe this needs one more indentation?

hiranya911: Go code is auto formatted by the `gofmt` tool, and this is how it has decided to format it. We do have a check in our CI that verifies correct formatting too. Given that we don't have a build break, I think we're good here.

Issue #229, opened 2019-03-14
FR: realtime event listeners for Realtime Database


johndelavega:
Or provide some general info on how I may be able to code it myself.

Thanks
John

hiranya911:
You can use the streaming support available in the RTDB REST API to implement this: https://firebase.google.com/docs/reference/rest/database/#section-streaming

That's how we supported it in our Python SDK.

johndelavega:
Looks like something I could do. You can assign it to me.

As you mentioned in issue#35 to do this incrementally, I could try to get something working with very basic support first then we can take it from there.

hiranya911:
Thanks @johndelavega. I don't think our repos allow assigning issues to non-Googlers. But consider it assigned :)

As a first step, please outline what the public API of this feature would look like here. Once we settle on that, you can go ahead and try to implement it. I assume this API will return some sort of a channel that allows clients to receive and consume updates. It might not be a bad idea to align with the semantics of the Python API here. See the `listen()` method at https://firebase.google.com/docs/reference/admin/python/firebase_admin.db#reference

johndelavega:
Here are my suggestion for public API and does not include a channel, just a callback function, but we can probably create one that does include a channel. Internal code will use goroutines for the async nature of a listener. The empty interface arg is similar to existing public Get/Set API.

func (r *Ref) Listen(ctx context.Context, func(v interface{}) (*Event, error)
func (r *Event) Close() error

Ideally, we can call Listen from the same *Ref multiple times with different callback functions, returning unique *Event (or *EventListener) object used for closing http connection when done.

Need clarification for the use case of ListenerRegistration in Python API, is this optional?

hiranya911:
The `ListenerRegistration` is equivalent to what you call `EventListener`. 

Perhaps, a better way to do this in Go is to align with the Firestore listen API: 

https://godoc.org/cloud.google.com/go/firestore#DocumentRef.Snapshots
https://godoc.org/cloud.google.com/go/firestore#DocumentSnapshotIterator

When applied to RTDB, this would look like this:

```
func (r *Ref) Listen(ctx context.Context) (*SnapshotIterator, error)

func (i *SnapshotIterator) Next() (*Event, error)
func (i *SnapshotIterator) Stop()

type Event struct {
  Path string
  EventType int // this can be an enum type
}
func (e *Event) Unmarshal(v interface{}) error
```

Something along these lines? @bklimt do you have any suggestions?

[...]

PR #245, opened 2019-05-25
Initial commit - FR: realtime event listener for RTDB (#229)

johndelavega:
Here's a pull request against ```dev```

``` see cmd/listen/main.go for working simple test client app```

To run the test app, edit main.go for credential json file location "edit 1." and db url "edit 2."
then from command line terminal, cd to cmd/listen
```go run . init```  - creates snapshot user1/path1/key1:"init"  for listener testing
```go run .``` - starts listening
 start a new command line terminal, cd to cmd/listen
```go run . hey``` - this will trigger an event and should show up from other cmd line listener 

Here's the gist of the code:
```golang
  iter, err := ref.Listen(ctx)
  
  defer iter.Stop()
  
  go func() {
    for {
      event, err := iter.Next()

      if err != nil {
        break
      }
    }
  }()
```

FR https://github.com/firebase/firebase-admin-go/issues/229

### DIFF
--- a/db/event_ref.go
+++ b/db/event_ref.go
@@ -1,0 +1,207 @@
+// Copyright 2019 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package db
+
+import (
+	"bufio"
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+
+	"firebase.google.com/go/internal"
+)
+
+// Listen ...
+func (r *Ref) Listen(ctx context.Context) (*SnapshotIterator, error) {
+
+	sseDataChan := make(chan string) // server-sent event data channel
+
+	var opts []internal.HTTPOption
+	opts = append(opts, internal.WithHeader("Cache-Control", "no-cache"))
+	opts = append(opts, internal.WithHeader("Accept", "text/event-stream"))
+	opts = append(opts, internal.WithHeader("Connection", "keep-alive"))
+
+	resp, err := r.sendListen(ctx, "GET", opts...)
+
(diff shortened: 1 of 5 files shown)


hiranya911:
Thanks @johndelavega. The general shape of things look good. However the SSE sessions will die after an hour, when the underlying credential becomes invalid. You will need a way to reconnect when this happens. Also try to use the APIs in the `json` package as much as possible to parse SSE data.

johndelavega:
Thanks for the feedback. Will work on the fix to reconnect. Will use `json` package for the SSE data parsing as that should make the code more resilient. Actually had plan to do that, so it was great seeing that feedback :)

johndelavega:
Here's a fix for the timeout reconnection. 

The code is tested and continued running for over 2 days.

The tested use case is similar to if we have a `client-go` SDK used for an IoT device or installed app using `"authorized_user"` JSON credential file, 
like this: [refresh_token.json](https://github.com/firebase/firebase-admin-go/blob/master/testdata/refresh_token.json)


The JSON file is created from `Google APIs` Credential `OAuth 2.0 Client IDs` using Type: `Other` 

This [tool create-refresh-token](https://github.com/johndelavega/create-refresh-token) can be used to create the JSON file.


This admin SDK can also be used as a client SDK, similar to Node.js SDK

johndelavega:
Updated to use `json` package APIs for parsing SSE data.

This `firebase-admin-go` SDK can be used as `firebase-client-go` SDK  using refresh_token `"authorized_user"` credential JSON file.

db/event_ref.go
```diff
@@ -0,0 +1,212 @@
...
+
+		if "event: put" == string(b) {
+
+			if scanner.Scan() == true {
+				b = scanner.Bytes()
+				s := string(b)
+
+				var snapshotMap map[string]interface{}
+
+				err := json.Unmarshal([]byte(s[6:]), &snapshotMap)
```
thatreguy: Better to comment why you've used a constant here

johndelavega: Thanks. I've added comments from a recent commit.

johndelavega: @thatreguy I think you need to sign a CLA https://cla.developers.google.com/

smoya:
Any update on this?

juicyJayy:
Any updates on this ?
Got the Event Listeners working thanks to @johndelavega  github repo.
Only issue I find is that the event is only triggered when I do manual changes from the firebase console.
However the event is NOT triggered when I change data from another app (flutter), nor is it triggered when I update data via the Firebase Database REST API.

iamkhalidbashir:
I think they are not going to work on this, sad.

Issue #213, opened 2018-12-27
FR: Sending message to multi devices


canhlinh:
Hi team,

Do you have any plan to support sending message to multiple devices ?
As technical document was described at here https://firebase.google.com/docs/cloud-messaging/android/send-multiple

That would be nice if the `firebase-go` can support this feature.

canhlinh:
@hiranya911 Sending message to multi tokens at the same time is a good choice.
I have already it with GCM and APNs. Please support on FCM

chong-shao:
Hi @hiranya911, I think currently multicast is not supported by go SDK. Is there anything I can help?

hiranya911:
I will eventually get to it. Developers are also invited to provide a pull request and help things move along.

PR #257, opened 2019-07-15
Implementing the messaging.SendAll() API

hiranya911:
Implemented the `SendAll()` API for sending up to 100 messages in a single RPC call. Also introducing the `SendResponse` and `BatchResponse` types.

go/fcm-multicast

Related to #213

### DIFF
--- a/messaging/messaging_batch.go
+++ b/messaging/messaging_batch.go
@@ -56,43 +56,23 @@ type BatchResponse struct {
 // this function indicates a total failure -- i.e. none of the messages in the list could be sent.
 // Partial failures are indicated by a `BatchResponse` return value.
 func (c *Client) SendAll(ctx context.Context, messages []*Message) (*BatchResponse, error) {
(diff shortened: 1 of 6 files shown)


messaging/messaging_batch.go
```diff
@@ -0,0 +1,273 @@
...
+
+// multipartEntity represents an HTTP entity that consists of multiple HTTP requests (parts).
+type multipartEntity struct {
+	parts []*part
+}
+
+func (c *Client) newBatchRequest(messages []*Message, dryRun bool) (*internal.Request, error) {
+	url := fmt.Sprintf("%s/projects/%s/messages:send", c.fcmEndpoint, c.project)
+	headers := map[string]string{
+		"X-GOOG-API-FORMAT-VERSION": "2",
```
bklimt: Should this be in a constant somewhere? It seems like the kind of header multiple APIs would use.

hiranya911: Done

Issue #259, opened 2019-07-19
FR: add image option in notification


chemidy:
There are new options for image in notification : https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages?authuser=0#notification

PR #261, opened 2019-07-19
Image

chemidy:
- add image option in notification

#259

### DIFF
--- a/messaging/messaging.go
+++ b/messaging/messaging.go
@@ -142,7 +142,7 @@ type Message struct {
 	Android      *AndroidConfig    `json:"android,omitempty"`
 	Webpush      *WebpushConfig    `json:"webpush,omitempty"`
 	APNS         *APNSConfig       `json:"apns,omitempty"`
-	FcmOptions   *FcmOptions       `json:"fcm_options,omitempty"`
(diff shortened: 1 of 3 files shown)


messaging/messaging_utils.go
```diff
@@ -58,6 +63,21 @@ func validateMessage(message *Message) error {
...
 }
 
+func validateNotification(notification *Notification) error {
+	if notification == nil {
+		return nil
+	}
+
+	if notification.Image != "" {
+		image := notification.Image
+		_, err := url.ParseRequestURI(image)
```
hiranya911: Nit: Inline the error check

```
if _, err := ....; err != nil {

}
```

Here and below

messaging/messaging.go
```diff
@@ -274,6 +275,7 @@ type AndroidNotification struct {
 	TitleLocKey  string   `json:"title_loc_key,omitempty"`
 	TitleLocArgs []string `json:"title_loc_args,omitempty"`
 	ChannelID    string   `json:"channel_id,omitempty"`
+	Image        string   `json:"image,omitempty"`
```
hiranya911: Lets rename to ImageURL to be consistent with the nomenclature used in Java SDK.

PR #265, opened 2019-07-31
Added snippets for new multicast APIs

hiranya911:
Snippets to appear on https://firebase.google.com/docs/cloud-messaging/send-message#send_messages_to_multiple_devices

### DIFF
--- a/snippets/messaging.go
+++ b/snippets/messaging.go
@@ -194,7 +194,7 @@ func sendMulticastAndHandleErrors(ctx context.Context, client *messaging.Client)
 		var failedTokens []string
 		for idx, resp := range br.Responses {
 			if !resp.Success {
-				// The order of responses corresponds to the order of the registration tokens
+				// The order of responses corresponds to the order of the registration tokens.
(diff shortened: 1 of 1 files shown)


snippets/messaging.go
```diff
@@ -136,8 +136,72 @@ func sendAll(ctx context.Context, client *messaging.Client) {
...
+	br, err := client.SendMulticast(context.Background(), message)
+	if err != nil {
+		log.Fatalln(err)
+	}
+
+	if br.FailureCount > 0 {
+		var failedTokens []string
+		for idx, resp := range br.Responses {
+			if !resp.Success {
+				// The order of responses corresponds to the order of the registration tokens
```
bklimt: 1. You might want to double check that this line isn't too long to look right on the website. A lot of the other comments are broken up to make short lines.

2. All of the other sentences in comments in this doc end with a period, so I'd recommend adding one here for consistency.

hiranya911: 1. Line length looks ok (87 columns). It's about the same length in case of the Java example.
2. Done.

PR #271, opened 2019-09-04
messaging: add image field to Notification 

trilopin:
Firebase has added a new field `Image` in their `Notification` spec and it seems to be the new way to pass images to notifications https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages?authuser=0#notification

Unfortunately, `Notification` struct has not included this change yet and the old way of sending images seems to not work anymore.

 I've added the new field to Notification struct. unit tests are passing, and this change includes the image into the push notification :).

Thanks!


trilopin:
@googlebot I signed it!

PR #274, opened 2019-09-12
Inline errors check

chemidy:
Nit: Inline errors check

```
if _, err := ....; err != nil {

}
```


messaging/messaging.go
```diff
@@ -262,18 +262,45 @@ func (a *AndroidConfig) UnmarshalJSON(b []byte) error {
...
-	Color        string   `json:"color,omitempty"` // notification color in #RRGGBB format
-	Sound        string   `json:"sound,omitempty"`
-	Tag          string   `json:"tag,omitempty"`
-	ClickAction  string   `json:"click_action,omitempty"`
-	BodyLocKey   string   `json:"body_loc_key,omitempty"`
-	BodyLocArgs  []string `json:"body_loc_args,omitempty"`
-	TitleLocKey  string   `json:"title_loc_key,omitempty"`
-	TitleLocArgs []string `json:"title_loc_args,omitempty"`
-	ChannelID    string   `json:"channel_id,omitempty"`
+	Title                 string           `json:"title,omitempty"` // if specified, overrides the Title field of the Notification type
```
hiranya911: Can we revert these changes? These really should be done as a separate pull request.

@lahirumaramba has an ongoing API proposal for this too.

PR #276, opened 2019-09-19
Use client context in verifyIDToken snippet

### LINKED ISSUE #268: Why is the client context not used in verifyIDToken?
For some reason the `ctx` that is passed in is not used, instead a new blank context is created for no purpose.

https://github.com/firebase/firebase-admin-go/blob/79d5dcdeff7f536eb11a549dd25579d88831d19e/snippets/auth.go#L80-L82

lahirumaramba:
- Use client context in `verifyIDToken` snippet

Resolves #268


hiranya911:
Please change the base branch to dev

PR #277, opened 2019-09-27
Remove (base64) 'REDACTED' passwords from user records.

rsgowman:
These values look like passwords hashes, but aren't, leading to
potential confusion.

Additionally, added docs to CONTRIBUTING.md detailing how to add the
permission that causes password hashes to be properly returned as well
as adjusting the test failure message should the developer not add that
permission.

b/141189502

### DIFF
--- a/auth/user_mgt_test.go
+++ b/auth/user_mgt_test.go
@@ -995,7 +995,7 @@ func TestInvalidDeleteUser(t *testing.T) {
 }
 
 func TestMakeExportedUser(t *testing.T) {
-	rur := &userQueryResponse{
+	queryResponse := &userQueryResponse{
 		UID:                "testuser",
 		Email:              "testuser@example.com",
 		PhoneNumber:        "+1234567890",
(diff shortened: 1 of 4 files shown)


auth/user_mgt_test.go
```diff
@@ -995,7 +995,7 @@ func TestInvalidDeleteUser(t *testing.T) {
 }
 
 func TestMakeExportedUser(t *testing.T) {
-	rur := &userQueryResponse{
+	queryResponse := &userQueryResponse{
```
rsgowman: I couldn't figure out what `rur` stood for, so I renamed it. (I wanted my test to be consistent with this one.) Happy to change it back if someone can enlighten me.

integration/auth/user_mgt_test.go
```diff
@@ -133,7 +133,16 @@ func TestListUsers(t *testing.T) {
 		if _, ok := newUsers[u.UID]; ok {
 			count++
 			if u.PasswordHash == "" {
-				t.Errorf("Users() PasswordHash = empty; want = non-empty")
+				t.Errorf(
+					"Users() PasswordHash = empty; want = non-empty. A common cause would be " +
```
hiranya911: Nit: Extract the "A common cause..." part into a constant.

rsgowman: Personally, I (slightly) prefer inlining it here. But I've already refactored in c# (due to the linter not liking multi-line parameters) and in python (due to 4x repetition) so at this point, refactoring everywhere *is* starting to feel more natural. Done.

PR #280, opened 2019-10-10
Added Create and Update APIs for SAML configs

hiranya911:
Added `CreateSAMLProviderConfig()` and `UpdateSAMLProviderConfig()` functions.

Also introducing a `nestedMap` helper type that makes it easier to model the create/update parameters as a nested map and build update masks.

### DIFF
--- a/auth/provider_config.go
+++ b/auth/provider_config.go
@@ -133,20 +133,17 @@ func (config *SAMLProviderConfigToCreate) ID(id string) *SAMLProviderConfigToCre
 
 // IDPEntityID sets the IDPEntityID field of the new config.
 func (config *SAMLProviderConfigToCreate) IDPEntityID(entityID string) *SAMLProviderConfigToCreate {
-	config.set(idpEntityIDKey, entityID)
-	return config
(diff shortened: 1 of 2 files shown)


auth/provider_config.go
```diff
@@ -40,6 +119,231 @@ type SAMLProviderConfig struct {
...
+// IDPEntityID sets the IDPEntityID field of the new config.
+func (config *SAMLProviderConfigToCreate) IDPEntityID(entityID string) *SAMLProviderConfigToCreate {
+	config.set(idpEntityIDKey, entityID)
+	return config
+}
+
+// SSOURL sets the SSOURL field of the new config.
+func (config *SAMLProviderConfigToCreate) SSOURL(url string) *SAMLProviderConfigToCreate {
+	config.set(ssoURLKey, url)
+	return config
```
lahirumaramba: Very minor thing... some places in the code we set the values first and then return the `config` object and in other places we do a single line return `return config.set(ssoURLKey, url)`. It might be better to keep it consistent everywhere

hiranya911: Good catch. Fixed.

PR #289, opened 2019-10-30
Bumped version to 3.10.0

lahirumaramba:
- Bumped version to 3.10.0


lahirumaramba:
@hiranya911 this includes your latest PR #288. Since it is not exposed through the API yet, I assumed it is okay to include it here. Please confirm. Thanks!

PR #292, opened 2019-11-01
Reject rounds=0 for SHA1 hashes

rsgowman:
Port of https://github.com/firebase/firebase-admin-node/pull/677


auth/hash/hash.go
```diff
@@ -240,8 +241,23 @@ func hmacConfig(name string, key []byte) (internal.HashConfig, error) {
...
 
 func basicConfig(name string, rounds int) (internal.HashConfig, error) {
-	if rounds < 0 || rounds > 120000 {
-		return nil, errors.New("rounds must be between 0 and 120000")
+	minRounds := 0
+	maxRounds := 120000
+	switch name {
+	case "MD5":
+		maxRounds = 8192
+	case "SHA1":
```
hiranya911: You can combine the cases as follows:

```
case "SHA1", "SHA256", "SHA512":
  minRounds = 1
  maxRounds = 8291
```

rsgowman: Done.

PR #293, opened 2019-11-05
fix: add go.mod

shogo82148:
From Go 1.13, module mode is the default.
`go.mod` file should be at the project root.
ref. https://blog.golang.org/using-go-modules

RELEASE NOTE: Added a `go.mod` file to the project marking it as a Go module.


shogo82148:
@googlebot I signed it!

shogo82148:
Thank you for your review. I fixed it.
Please review it again!

PR #297, opened 2019-11-06
fix(fcm): Increased FCM batch request limit to 500

hiranya911:
RELEASE NOTE: Batch messaging APIs `SendAll()` and `SendMulticast()` now support sending up to 500 messages in a single call.


hiranya911:
It was an Emacs crash recovery file that was checked in accidentally. Removed. Thanks for pointing it out.

PR #304, opened 2019-11-14
Added Tenants() function to iterate over tenants

hiranya911:
API for iterating over all tenants. Based on the google.golang.org/api/iterator framework.

### DIFF
--- a/auth/tenant_mgt.go
+++ b/auth/tenant_mgt.go
@@ -19,9 +19,11 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 
 	"firebase.google.com/go/internal"
+	"google.golang.org/api/iterator"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


auth/tenant_mgt.go
```diff
@@ -198,6 +200,33 @@ func (tm *TenantManager) DeleteTenant(ctx context.Context, tenantID string) erro
...
+// iterator starts after the token.
+func (tm *TenantManager) Tenants(ctx context.Context, nextPageToken string) *TenantIterator {
+	it := &TenantIterator{
+		ctx: ctx,
+		tm:  tm,
+	}
+	it.pageInfo, it.nextFunc = iterator.NewPageInfo(
+		it.fetch,
+		func() int { return len(it.tenants) },
+		func() interface{} { b := it.tenants; it.tenants = nil; return b })
```
lahirumaramba: Nit: Should the closing `)` be in a new line?

hiranya911: We've used this style in other files (provider_config.go, user_mgt.go). So my preference would be to keep it this way.

PR #316, opened 2019-12-18
feat(fcm): Introduced a series of new parameters to `AndroidNotification` type.

### LINKED ISSUE #273: FR : add new options to AndroidNotification
There are new options for AndroidNotification : https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages?authuser=0#androidnotification

> "ticker": string,
  "sticky": boolean,
  "event_time": string,
  "local_only": boolean,
  "notification_priority": enum (NotificationPriority),
  "default_sound": boolean,
  "default_vibrate_timings": boolean,
  "default_light_settings": boolean,
  "vibrate_timings": [
    string
  ],
  "visibility": enum (Visibility),
  "notification_count": number,
  "light_settings": {
    object (LightSettings)
  },

### LINKED ISSUE #308: Missing notification_priority  in Go SDK
**Environment Details**
  * Operating System version: WIndows 10 
  * Firebase SDK version: 3.10.0
  * Firebase Product: Firebase Cloud Messaging

As Per firebase document, we can set the parameter **notification_priority** to determine message priority.

https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#androidnotification

  ```
  {
      "title": string,
      "body": string,
      "icon": string,
      "color": string,
      "sound": string,
      "tag": string,
      "click_action": string,
      "body_loc_key": string,
      "body_loc_args": [
        string
      ],
      "title_loc_key": string,
      "title_loc_args": [
        string
      ],
      "channel_id": string,
      "ticker": string,
      "sticky": boolean,
      "event_time": string,
      "local_only": boolean,
      "notification_priority": enum (NotificationPriority),

```
I am trying to use firebase admin go client but when I see the message struct, I cannot see that element 

Struct Defined is as follows:
https://godoc.org/firebase.google.com/go/messaging#AndroidNotification

   ```
 type AndroidNotification struct {
        Title        string   `json:"title,omitempty"` // if specified, overrides the Title field of the Notification type
        Body         string   `json:"body,omitempty"`  // if specified, overrides the Body field of the Notification type
        Icon         string   `json:"icon,omitempty"`
        Color        string   `json:"color,omitempty"` // notification color in #RRGGBB format
        Sound        string   `json:"sound,omitempty"`
        Tag          string   `json:"tag,omitempty"`
        ClickAction  string   `json:"click_action,omitempty"`
        BodyLocKey   string   `json:"body_loc_key,omitempty"`
        BodyLocArgs  []string `json:"body_loc_args,omitempty"`
        TitleLocKey  string   `json:"title_loc_key,omitempty"`
        TitleLocArgs []string `json:"title_loc_args,omitempty"`
        ChannelID    string   `json:"channel_id,omitempty"`
        ImageURL     string   `json:"image,omitempty"`
    }
```

Is there any discrepancy in firebase documentation and struct or am I missing something?
Issue is, message are delivered to device but no sound or on screen display if device is sleeping. ( Happening on one plus 7 pro, only sometimes)

I checked the events using following commands 

       1.  Dial *#*#426#*#* to open the FCM Diagnostics page.
       2.  Tap the EVENTS button to show the event log.
       3.  Check whether the message was delivered to the device. There should be a log entry reading 

and saw that youtube message are coming with **HIGH** priority and priority of my messages is shown as **UNKNOWN**.

How I can send message with high priority as described in documentation?

hiranya911:
Resolves #308 
Resolves #273 

I deviated slightly from go/admin-sdk-fcm-12-new-parameters-node-plus by taking Priority and Visibility parameters as values instead of pointers. This makes the calling code a lot simpler to the developer. Passing pointers to constants is rather tedious in Go.

### DIFF
--- a/messaging/messaging.go
+++ b/messaging/messaging.go
@@ -367,7 +367,7 @@ type AndroidNotificationPriority int
 const (
 	priorityUnspecified AndroidNotificationPriority = iota
 
-	// PriorityMin is the owest notification priority. Notifications with this priority might not
+	// PriorityMin is the lowest notification priority. Notifications with this priority might not
 	// be shown to the user except under special circumstances, such as detailed notification logs.
 	PriorityMin
 
@@ -514,11 +514,19 @@ func newColor(clr string) (*color, error) {
 		return nil, fmt.Errorf("failed to parse %s: %v", clr, err)
 	}
 
+	alpha := int64(255)
+	if len(clr) == 9 {
+		alpha, err = strconv.ParseInt(clr[7:9], 16, 32)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse %s: %v", clr, err)
+		}
+	}
+
 	return &color{
 		Red:   float64(red) / 255.0,
 		Green: float64(green) / 255.0,
 		Blue:  float64(blue) / 255.0,
-		Alpha: 1.0,
+		Alpha: float64(alpha) / 255.0,
 	}, nil
 }
 
@@ -526,7 +534,11 @@ func (c *color) toString() string {
 	red := int(c.Red * 255.0)
 	green := int(c.Green * 255.0)
 	blue := int(c.Blue * 255.0)
-	return fmt.Sprintf("#%X%X%X", red, green, blue)
+	alpha := int(c.Alpha * 255.0)
+	if alpha == 255 {
+		return fmt.Sprintf("#%X%X%X", red, green, blue)
+	}
+	return fmt.Sprintf("#%X%X%X%X", red, green, blue, alpha)
 }
 
 // AndroidFCMOptions contains additional options for features provided by the FCM Android SDK.
--- a/messaging/messaging.go
+++ b/messaging/messaging.go
@@ -48,6 +48,8 @@ const (
 	serverUnavailable              = "server-unavailable"
 	tooManyTopics                  = "too-many-topics"
 	unknownError                   = "unknown-error"
+
+	rfc3339Zulu = "2006-01-02T15:04:05.000000000Z"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


messaging/messaging.go
```diff
@@ -209,42 +205,328 @@ func (a *AndroidConfig) UnmarshalJSON(b []byte) error {
...
+		}
+		if vis, ok := visibilities[temp.Visibility]; ok {
+			a.Visibility = vis
+		} else {
+			return fmt.Errorf("unknown visibility value: %q", temp.Visibility)
+		}
+	}
+
+	if temp.EventTimestamp != "" {
+		ts, err := time.Parse(rfc3339Zulu, temp.EventTimestamp)
```
lahirumaramba: Is `time.Parse(...)` aware that this timestamp is in UTC? Looking at the docs I think this converts the time to the local timezone, which is what we need here... but I am not sure if the function is aware that the input timestamp is in UTC

messaging/messaging.go
```diff
@@ -209,42 +205,328 @@ func (a *AndroidConfig) UnmarshalJSON(b []byte) error {
...
+	return nil
+}
+
+// AndroidNotificationPriority represents the priority levels of a notification.
+type AndroidNotificationPriority int
+
+const (
+	priorityUnspecified AndroidNotificationPriority = iota
+
+	// PriorityMin is the owest notification priority. Notifications with this priority might not
```
lahirumaramba: Nit:
```suggestion
	// PriorityMin is the lowest notification priority. Notifications with this priority might not
```

messaging/messaging.go
```diff
@@ -209,42 +205,328 @@ func (a *AndroidConfig) UnmarshalJSON(b []byte) error {
...
+	blue, err := strconv.ParseInt(clr[5:7], 16, 32)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse %s: %v", clr, err)
+	}
+
+	return &color{
+		Red:   float64(red) / 255.0,
+		Green: float64(green) / 255.0,
+		Blue:  float64(blue) / 255.0,
+		Alpha: 1.0,
```
lahirumaramba: I think you can use the `Alpha` channel to control the brightness of the LED. Should we support it in the API?
We could add it in a future PR, too

messaging/messaging.go
hiranya911: Yes. According to https://golang.org/pkg/time/#Parse

> In the absence of a time zone indicator, Parse returns a time in UTC.

messaging/messaging.go

messaging/messaging.go
hiranya911: Added

hiranya911:
@lahirumaramba please take another look. I've added support for color strings with alpha values.

PR #318, opened 2019-12-20
Migrate a few tests to use https://github.com/stretchr/testify

rsgowman:


rsgowman:
wdyt? Better or worse?

hiranya911:
I'm going to side with the guidelines provided in go/gocomments#assert here. Basically we prefer Go's built-in error checking and reporting functions over assert libraries.

PR #325, opened 2020-01-15
Add bulk get/delete methods

### LINKED ISSUE #138: FR: Support getting multiple user records at once by email

rsgowman:
This PR allows callers to retrieve a list of users by unique identifier (uid, email, phone, federated provider uid) as well as to delete a list of users.

Resolves #138


auth/user_mgt.go
```diff
@@ -561,6 +582,261 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
 	return parsed.Users[0].makeUserRecord()
 }
 
+// A UserIdentifier identifies a user to be looked up.
+type UserIdentifier interface {
+	String() string
```
hiranya911: I don't think this should be in the public API. Call is something like `toString()`.

auth/user_mgt.go
```diff
@@ -561,6 +582,261 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+
+func (id UIDIdentifier) matchesUserRecord(ur *UserRecord) (bool, error) {
+	err := id.validate()
+	if err != nil {
+		return false, err
+	}
+	if id.UID == ur.UID {
+		return true, nil
+	}
+	return false, nil
```
hiranya911:
```
return id.UID == ur.UID, nil
```

auth/user_mgt.go
```diff
@@ -561,6 +582,261 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+}
+
+func (id UIDIdentifier) populateRequest(req *getAccountInfoRequest) error {
+	err := id.validate()
```
hiranya911: I think it's better to have a single `validate()` method on `getAccountIntoRequest`. Once it has been fully populated, you can call `req.validate()` in the `GetUsers()` function.

auth/user_mgt.go
```diff
@@ -561,6 +582,261 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
 }
 
+// A UserIdentifier identifies a user to be looked up.
+type UserIdentifier interface {
+	String() string
+
+	validate() error
+	matchesUserRecord(ur *UserRecord) (bool, error)
+	populateRequest(req *getAccountInfoRequest) error
+}
```
hiranya911: Interfaces with many functions (> 2) are a code smell in Go. I think we can push most of this logic onto `getAccountInfoRequest`. See my other comments. Hopefully we can bring it down to just one function: `populateRequest()`

auth/user_mgt.go
```diff
@@ -561,6 +582,261 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+
+func (id UIDIdentifier) String() string {
+	return fmt.Sprintf("UIDIdentifier{%s}", id.UID)
+}
+
+func (id UIDIdentifier) validate() error {
+	return validateUID(id.UID)
+}
+
+func (id UIDIdentifier) matchesUserRecord(ur *UserRecord) (bool, error) {
```
hiranya911: We can move this logic to `getAccountInfoRequest`. In pseudo code:

```
req := InitGetAccountInfoRequest()
Populate(req, uids)
result := MakeRPC(req)
for _, uid := range req.localId {
  // Search for missing UIDs
}
for _, email := range req.email {
  // Search for missing emails
}
// And so on for phone and federatedIds
```

This looks a bit crass, but it does have a couple of advantages:
1. We don't have to validate the identifiers after the RPC has been made. Since we are not looking at the user arguments once the request has been created, we know it hasn't been modified by the user.
2. We can remove `validate()` and `matchesUserRecord()` from this interface.

auth/user_mgt.go
```diff
@@ -561,6 +582,261 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+
+func (req *getAccountInfoRequest) build() map[string]interface{} {
+	var builtFederatedUserID []map[string]interface{}
+	for i := range req.federatedUserID {
+		builtFederatedUserID = append(builtFederatedUserID, map[string]interface{}{
+			"providerId": req.federatedUserID[i].providerID,
+			"rawId":      req.federatedUserID[i].rawID,
+		})
+	}
+	return map[string]interface{}{
```
hiranya911: Why not just add the `json: ...` annotations to the fields, and let Go's `json` package do its thing?

auth/user_mgt.go
```diff
@@ -561,6 +582,261 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+	// Set of UserIdentifiers that were requested, but not found.
+	NotFound []UserIdentifier
+}
+
+type federatedUserIdentifier struct {
+	providerID string
+	rawID      string
+}
+
+type getAccountInfoRequest struct {
```
hiranya911: Implement a validation function for this type. Then in GetUsers:

```
if err := req.validate(); err != nil {
  return nil, err
}

_, err := c.post(ctx, "/accounts:lookup", request, &parsed)
```

auth/user_mgt.go
```diff
@@ -561,6 +582,261 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+		return &GetUsersResult{[](*UserRecord){}, [](UserIdentifier){}}, nil
+	} else if len(identifiers) > maxGetAccountsBatchSize {
+		return nil, internal.Errorf(
+			maximumUserCountExceeded,
+			"`identifiers` parameter must have <= %d entries.", maxGetAccountsBatchSize)
+	}
+
+	var request getAccountInfoRequest
+	for i := range identifiers {
+		err := identifiers[i].populateRequest(&request)
```
hiranya911: Use the in-lined syntax. Here and elsewhere:

```
if err := ....; err != nil {

}
```

auth/user_mgt.go
```diff
@@ -728,6 +1015,101 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
...
+		Index   int    `json:"index"`
+		LocalID string `json:"localId"`
+		Message string `json:"message"`
+	}
+	type batchDeleteAccountsResponse struct {
+		Errors []batchDeleteErrorInfo `json:"errors"`
+	}
+
+	resp := batchDeleteAccountsResponse{}
+	_, err := c.post(ctx, "/accounts:batchDelete", payload, &resp)
```
hiranya911: Return early on error.

auth/user_mgt.go
```diff
@@ -728,6 +1015,101 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
...
+
+	for i := range uids {
+		if err := validateUID(uids[i]); err != nil {
+			return nil, err
+		}
+
+		payload.LocalIds = append(payload.LocalIds, uids[i])
+	}
+
+	type batchDeleteErrorInfo struct {
```
hiranya911: Don't think you need this type. Add the json annotations to the public `DeleteUsersErrorInfo` type and unmarshal into it directly.

auth/user_mgt.go
```diff
@@ -728,6 +1015,101 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
...
+
+	resp := batchDeleteAccountsResponse{}
+	_, err := c.post(ctx, "/accounts:batchDelete", payload, &resp)
+
+	result := DeleteUsersResult{
+		FailureCount: len(resp.Errors),
+		SuccessCount: len(uids) - len(resp.Errors),
+	}
+
+	for i := range resp.Errors {
```
hiranya911: You won't need this if you directly unmarshal into `DeleteUsersErrorInfo`. Then this will just become:

```
result.Errors = resp.Errors
```

auth/user_mgt_test.go
```diff
@@ -157,6 +158,267 @@ func TestInvalidGetUser(t *testing.T) {
...
+		if users[i].UID != uids[i] {
+			return false
+		}
+	}
+
+	return true
+}
+
+func TestGetUsers(t *testing.T) {
+	t.Run("should be rejected when given more than 100 identifiers", func(t *testing.T) {
```
hiranya911: Each of these `t.Run` calls should be a top-level test case. Just give them descriptive names:

```
TestGetUsersExceeds100
TestGetUsersEmpty
TestGetUsersAllNonExisting
etc...
```

Sub tests are useful when you want to share some state between two or more test cases.

auth/user_mgt.go
```diff
@@ -561,6 +582,261 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+	return false, nil
+}
+
+func (c *baseClient) GetUsers(
+	ctx context.Context, identifiers []UserIdentifier,
+) (*GetUsersResult, error) {
+	if len(identifiers) == 0 {
+		return &GetUsersResult{[](*UserRecord){}, [](UserIdentifier){}}, nil
+	} else if len(identifiers) > maxGetAccountsBatchSize {
+		return nil, internal.Errorf(
```
hiranya911: We don't use error codes for developer errors. Just return a normal error. Use error codes for back-end errors only.

auth/user_mgt_test.go
```diff
@@ -157,6 +158,267 @@ func TestInvalidGetUser(t *testing.T) {
...
+	})
+
+	t.Run("should return no results when given no identifiers", func(t *testing.T) {
+		client := &Client{
+			baseClient: &baseClient{},
+		}
+
+		getUsersResult, err := client.GetUsers(context.Background(), [](UserIdentifier){})
+		if getUsersResult == nil || err != nil {
+			t.Errorf("GetUsers([]) failed with: %s", err.Error())
```
hiranya911: Do `t.Fatalf` here and drop the `else`:

```
if error_cond {
  t.Fatalf(...)
}

// Remaining validation logic
```

auth/user_mgt.go
rsgowman: (The validate() thing is no longer applicable.)

I'm not sure this is a worthwhile trade-off. Without the extra interface method (matchesUserRecord) we have to test all possible "sub-types". If we add a new subtype, we won't have any compile time safety to ensure the check is added here.

Additionally, we'll have to check to see if each type is either a (eg) UIDIdentifier or a *UIDIdentifier. I don't know a way to do that without duplicating the code. (But I don't know go very well, so suggestions are welcome.)

I've gone ahead and done this anyways, but in a separate commit. I'd prefer to just drop the commit, but if you prefer it (possibly with some improvements), then we can include it.

If we must get down to <=2 methods in the interface, I'd prefer to drop toString(). It's not currently used anywhere, and since it's not private, it can't be used by the end user either. (In fact, I think I'd like to drop it regardless of this issue; see my other comment.)

auth/user_mgt.go
rsgowman: (See my other comment for my response.)

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,258 @@ func TestGetNonExistingUser(t *testing.T) {
...
+			client.DeleteUser(context.Background(), createdUsersUids[i])
+		}
+	}
+
+	// Checks to see if the users list contain the given uids. Order is ignored.
+	//
+	// Behaviour is undefined if there are duplicate entries in either of the
+	// slices.
+	//
+	// This function is identical to the one in integration/auth/user_mgt_test.go
```
hiranya911: Comment not needed here.

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,258 @@ func TestGetNonExistingUser(t *testing.T) {
...
+		Email("user1@example.com").
+		PhoneNumber("+15555550001")
+	testUser2 := (&auth.UserToCreate{}).
+		UID("uid2").
+		Email("user2@example.com").
+		PhoneNumber("+15555550002")
+	testUser3 := (&auth.UserToCreate{}).
+		UID("uid3").
+		Email("user3@example.com").
+		PhoneNumber("+15555550003")
```
hiranya911: Use `newUserWithParams()` to create the first 3 users. Then you can get rid of the `createUser()` helper function.

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,258 @@ func TestGetNonExistingUser(t *testing.T) {
...
+
+	importUser1 := (&auth.UserToImport{}).
+		UID("uid4").
```
hiranya911: Use the random ID generator functions `randomUID()`, `randomEmail()` etc

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,258 @@ func TestGetNonExistingUser(t *testing.T) {
...
+				err = fmt.Errorf("Unable to create user %v: %v", *userToCreate, err)
+				return nil
+			}
+			return userRecord
+		}
+
+		// Helper to import a user and return its UserRecord. Upon error, sets the
+		// err variable. `uid` must match the UID set on the `userToImport`
+		// parameter.
+		importUser := func(uid string, userToImport *auth.UserToImport) *auth.UserRecord {
```
hiranya911: Move to the top level of the package.

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,258 @@ func TestGetNonExistingUser(t *testing.T) {
...
+		createUser(testUser3)
+
+		importUser("uid4", importUser1)
+
+		return err
+	}
+
+	// Delete all test users. This function will attempt to delete all test users
+	// even if a failure occurs.
+	deleteTestUsers := func() {
```
hiranya911: Can be replaced with:

```
defer deleteUser(testUser1.UID)
defer deleteUser(testUser2.UID)
// etc
```

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,258 @@ func TestGetNonExistingUser(t *testing.T) {
...
+				return false
+			}
+		}
+
+		return true
+	}
+
+	// Delete all the users that we're about to create (in case they were left
+	// over from a prior run).
+	deleteTestUsers()
```
hiranya911: Once you switch to randomly generated IDs, this can be removed.

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,258 @@ func TestGetNonExistingUser(t *testing.T) {
...
+			if getUsersResult.Users[0].UID != "uid1" {
+				t.Errorf("GetUsers([...]).Users[0].UID = %s; want = 'uid1'", getUsersResult.Users[0].UID)
+			}
+		}
+		if len(getUsersResult.NotFound) != 0 {
+			t.Errorf("len(GetUsers([...]).NotFound) = %d; want = 0", len(getUsersResult.NotFound))
+		}
+	})
+
+	t.Run("returns users with a LastRefreshTimestamp", func(t *testing.T) {
```
hiranya911: This is not really related to GetUsers() if I understood correct. May be define as a separate top-level test case. Or see if we can add the relevant checks to the existing `TestCreateUser`  test case.

integration/auth/user_mgt_test.go
```diff
@@ -334,6 +587,124 @@ func TestDeleteUser(t *testing.T) {
 	}
 }
 
+func TestDeleteUsers(t *testing.T) {
+	// Creates a user and returns its uid. Upon failure, triggers t.Fatalf().
+	createTestUserOrDie := func(t *testing.T) string {
```
hiranya911: Use the existing `newUserWithParams()` helper.

integration/auth/user_mgt_test.go
```diff
@@ -334,6 +587,124 @@ func TestDeleteUser(t *testing.T) {
...
+	// deleting the users to ensure the delete method worked.
+	ensureUsersNotFound := func(t *testing.T, uids []string) {
+		identifiers := []auth.UserIdentifier{}
+		for i := range uids {
+			identifiers = append(identifiers, auth.UIDIdentifier{UID: uids[i]})
+		}
+
+		getUsersResult, err := client.GetUsers(context.Background(), identifiers)
+		if err != nil {
+			t.Errorf("GetUsers(notfound_ids) error %v; want nil", err)
```
hiranya911: Fatalf and drop else

integration/auth/user_mgt_test.go
```diff
@@ -334,6 +587,124 @@ func TestDeleteUser(t *testing.T) {
...
+		} else {
+			ok := true
+			if len(getUsersResult.Users) != 0 {
```
hiranya911: I think this condition is redundant.

integration/auth/user_mgt_test.go
```diff
@@ -334,6 +587,124 @@ func TestDeleteUser(t *testing.T) {
...
+				t.Errorf("len(GetUsers(notfound_ids).Users) = %d; want 0", len(getUsersResult.Users))
+				ok = false
+			}
+			if len(getUsersResult.NotFound) != len(uids) {
+				t.Errorf("len(GetUsers(notfound_ids).NotFound) = %d; want %d", len(getUsersResult.NotFound), len(uids))
```
hiranya911: t.Fatalf and remove `ok`

integration/auth/user_mgt_test.go
rsgowman: Done.

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,219 @@ func TestGetNonExistingUser(t *testing.T) {
...
+		if err != nil {
+			t.Fatalf("GetUsers([...]) returned an error: %v", err)
+		}
+
+		if !sameUsers(getUsersResult.Users, []string{testUser1.UID, testUser3.UID}) {
+			t.Errorf("GetUsers([valid identifiers]) = %v; want = %v (in any order)",
+				getUsersResult.Users, []string{testUser1.UID, testUser3.UID})
+		}
+		if len(getUsersResult.NotFound) != 1 {
+			t.Errorf("len(GetUsers([...]).NotFound) = %d; want 1", len(getUsersResult.NotFound))
```
hiranya911: t.Fatalf?

integration/auth/user_mgt_test.go
hiranya911: Shouldn't this be removed?

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,219 @@ func TestGetNonExistingUser(t *testing.T) {
...
+
+	userRecordsToUIDs := func(users [](*auth.UserRecord)) []string {
+		results := []string{}
+		for i := range users {
+			results = append(results, users[i].UID)
+		}
+		return results
+	}
+
+	t.Run("returns users by various identifier types in a single call", func(t *testing.T) {
```
hiranya911: Lengthy test names/descriptions are not all that common in Go. Although I haven't found any concrete guidelines on the topic either. I'd say take a look at the tenant management integration tests, where we've used short names. But I'll leave it up to you to decide whether to use that pattern here or not.

auth/user_mgt_test.go
```diff
@@ -157,6 +158,265 @@ func TestInvalidGetUser(t *testing.T) {
...
+func TestGetUsersInvalidUid(t *testing.T) {
+	client := &Client{
+		baseClient: &baseClient{},
+	}
+
+	getUsersResult, err := client.GetUsers(
+		context.Background(),
+		[]UserIdentifier{&UIDIdentifier{"too long " + strings.Repeat(".", 128)}})
+	if getUsersResult != nil || err == nil {
+		t.Fatalf("GetUsers([too-long]) should fail")
```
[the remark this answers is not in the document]
rsgowman: Done.

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,219 @@ func TestGetNonExistingUser(t *testing.T) {
...
+		for i := range users {
+			if users[i].UID != uids[i] {
+				return false
+			}
+		}
+
+		return true
+	}
+
+	createTestUsers(t)
```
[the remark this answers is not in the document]
rsgowman: Done.

integration/auth/user_mgt_test.go
```diff
@@ -91,6 +92,219 @@ func TestGetNonExistingUser(t *testing.T) {
...
+
+	t.Run("returns users by various identifier types in a single call", func(t *testing.T) {
+		getUsersResult, err := client.GetUsers(context.Background(), []auth.UserIdentifier{
+			auth.UIDIdentifier{UID: testUser1.UID},
+			auth.EmailIdentifier{Email: testUser2.Email},
+			auth.PhoneIdentifier{PhoneNumber: testUser3.PhoneNumber},
+			auth.ProviderIdentifier{ProviderID: "google.com", ProviderUID: "google_" + importUser1UID},
+		})
+		if err != nil {
+			t.Fatalf("GetUsers([valid identifiers]) returned an error: %v", err)
```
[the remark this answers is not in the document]
rsgowman: Done.

auth/user_mgt.go
hiranya911: This is something we've been working on. Lets discuss some other time. go/firebase-error-handling if you need some extra context in the meantime.

integration/auth/user_mgt_test.go
rsgowman: This was deliberate (though also somewhat pedantic). I can clearly use Fatalf now to eliminate the else, but if the GetUsersResult struct gets an extra field added to it and a corresponding block added to this test, then the Fatalf would needlessly stop the test before that extra field could be verified.

integration/auth/user_mgt_test.go
rsgowman: Yup. :)

integration/auth/user_mgt_test.go
rsgowman: Those ones seem to be using the function name as the test name/description. (Which makes sense since in those cases, the SDK is effectively just calling the RPC. We're doing a little extra processing here.) I can't apply that directly, since these tests all use the same function name, but lets see if I can add some terseness.

auth/user_mgt.go
```diff
@@ -561,6 +581,187 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+	for i := range urs {
+		match := id.matchesUserRecord(urs[i])
+		if match {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *baseClient) GetUsers(
```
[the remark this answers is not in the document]
rsgowman: Done. (Yikes! Thanks for noticing!)

auth/user_mgt.go
```diff
@@ -728,6 +951,91 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
...
+// account.
+//
+// The Index field corresponds to the index of the failed user in the uids
+// array that was passed to DeleteUsers().
+type DeleteUsersErrorInfo struct {
+	Index  int    `json:"index,omitEmpty"`
+	Reason string `json:"message,omitEmpty"`
+}
+
+// Deletes the users specified by the given identifiers.
```
hiranya911: Same here.

auth/user_mgt.go
```diff
@@ -561,6 +581,199 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+func isUserFound(id UserIdentifier, urs [](*UserRecord)) bool {
+	for i := range urs {
+		if id.matches(urs[i]) {
+			return true
+		}
+	}
+	return false
+}
+
+// Gets the user data corresponding to the specified identifiers.
```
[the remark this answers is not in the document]
rsgowman: Done.

auth/user_mgt.go
rsgowman: Done.

auth/user_mgt.go
```diff
@@ -561,6 +581,195 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+
+func (id ProviderIdentifier) populate(req *getAccountInfoRequest) {
+	req.FederatedUserID = append(
+		req.FederatedUserID,
+		federatedUserIdentifier{ProviderID: id.ProviderID, RawID: id.ProviderUID})
+}
+
+// A GetUsersResult represents the result of the GetUsers() API.
+type GetUsersResult struct {
+	// Set of UserRecords, corresponding to the set of users that were requested.
```
egilmorez: Suggest omitting this comma.

auth/user_mgt.go
```diff
@@ -561,6 +581,195 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+	return false
+}
+
+// GetUsers returns the user data corresponding to the specified identifiers.
+//
+// There are no ordering guarantees; in particular, the nth entry in the users
+// result list is not guaranteed to correspond to the nth entry in the input
+// parameters list.
+//
+// Only a maximum of 100 identifiers may be supplied. If more than 100
```
egilmorez: Suggest "A maximum..."

auth/user_mgt.go
```diff
@@ -561,6 +581,195 @@ func (c *baseClient) getUser(ctx context.Context, query *userQuery) (*UserRecord
...
+// identifiers are supplied, this method will immediately return an error.
```
egilmorez: Suggest "method returns an error."

auth/user_mgt.go
```diff
@@ -728,6 +947,91 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
 	return err
 }
 
+// A DeleteUsersResult represents the result of the DeleteUsers() call.
+type DeleteUsersResult struct {
+	// The number of users that were deleted successfully (possibly zero). Users
+	// that did not exist prior to calling DeleteUsers() will be considered to be
```
egilmorez: Suggest "are considered"

auth/user_mgt.go
```diff
@@ -728,6 +947,91 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
...
+	// successfully deleted.
+	SuccessCount int
+
+	// The number of users that failed to be deleted (possibly zero).
+	FailureCount int
+
+	// A list of describing the errors that were encountered
```
egilmorez: Delete "of" if this list actually describes the errors.  Otherwise, just "A list of errors encountered..."

Suggest checking whether this was copied into similar comment in other SDKs, and we just missed it :)

auth/user_mgt.go
```diff
@@ -728,6 +947,91 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
...
+// array that was passed to DeleteUsers().
+type DeleteUsersErrorInfo struct {
+	Index  int    `json:"index,omitEmpty"`
+	Reason string `json:"message,omitEmpty"`
+}
+
+// DeleteUsers deletes the users specified by the given identifiers.
+//
+// Deleting a non-existing user won't generate an error. (i.e. this method is
+// idempotent.) Non-existing users will be considered to be successfully
```
egilmorez: Suggest "are considered"

auth/user_mgt.go
```diff
@@ -728,6 +947,91 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
...
+// deleted, and will therefore be counted in the DeleteUsersResult.SuccessCount
```
egilmorez: Suggest "are counted"

auth/user_mgt.go
```diff
@@ -728,6 +947,91 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
...
+// value.
+//
+// Only a maximum of 1000 identifiers may be supplied. If more than 1000
```
egilmorez: Suggest "A maximum..."

auth/user_mgt.go
```diff
@@ -728,6 +947,91 @@ func (c *baseClient) DeleteUser(ctx context.Context, uid string) error {
...
+// identifiers are supplied, this method will immediately return an error.
```
egilmorez: Suggest "method returns an error"

auth/user_mgt.go
rsgowman: Oh; the type somehow got dropped. (Which is mildly redundant info in most ports since you can get it from the type signature; python being the obvious exception.) The other ports seem ok.

I've added the type to match the other ports. Alternatively, we could just remove the type everywhere (except python).

auth/user_mgt.go
rsgowman: Done.

auth/user_mgt.go
rsgowman: Done.

auth/user_mgt.go
rsgowman: Done.

auth/user_mgt.go
rsgowman: Done.

auth/user_mgt.go
```diff
@@ -34,6 +34,12 @@ const (
 	maxLenPayloadCC     = 1000
 	defaultProviderID   = "firebase"
 	idToolkitV1Endpoint = "https://identitytoolkit.googleapis.com/v1"
+
+	// Maximum allowed number of users to batch get at one time.
+	maxGetAccountsBatchSize = 100
+
+	// Maximum allowed numberof users to batch delete at one time.
```
RelaxedPear: Add space between "number" and "of"

auth/user_mgt.go
```diff
@@ -34,6 +34,12 @@ const (
 	maxLenPayloadCC     = 1000
 	defaultProviderID   = "firebase"
 	idToolkitV1Endpoint = "https://identitytoolkit.googleapis.com/v1"
+
+	// Maximum allowed number of users to batch get at one time.
```
RelaxedPear: Suggest: "Maximum number of users allowed to batch get at a time."

auth/user_mgt.go
```diff
@@ -57,6 +63,9 @@ type UserInfo struct {
 type UserMetadata struct {
 	CreationTimestamp  int64
 	LastLogInTimestamp int64
+	// The time at which the user was last active (ID token refreshed), or 0 if
+	// the user was never active.
+	LastRefreshTimestamp int64
 }
 
 // UserRecord contains metadata associated with a Firebase user account.
```
RelaxedPear: `UserRecord`? Not familiar with the style guide for GO, please add back ticks if necessary.

auth/user_mgt.go
rsgowman: Done. I've also applied this to the maxDeleteAccountsBatchSize comment immediately below.

auth/user_mgt.go
rsgowman: Done.

auth/user_mgt.go
rsgowman: It doesn't seem to be a thing; or at least, they don't do it here: https://blog.golang.org/godoc-documenting-go-code or here: https://golang.org/doc/effective_go.html#commentary

Nonetheless, we do seem to use backticks occasionally... though seemingly only when referring to some other function (and I suspect inconsistently at that.)

I've left this alone.

.github/scripts/publish_post_check.sh
```diff
@@ -0,0 +1,105 @@
+
```
hiranya911, line 1: Let's remove this along with `stage.yml` and `publish.yml` from the PR. They are no longer in the dev branch.

auth/user_mgt.go
```diff
@@ -331,6 +340,7 @@ const (
 	unauthorizedContinueURI  = "unauthorized-continue-uri"
 	unknown                  = "unknown-error"
 	userNotFound             = "user-not-found"
+	maximumUserCountExceeded = "maximum-user-count-exceeded"
```
hiranya911, line 343: This shouldn't be needed. We don't use error codes for developer errors.

.github/scripts/publish_post_check.sh
rsgowman: Done.

auth/user_mgt.go
rsgowman: Done.

PR #328, opened 2020-02-04
Snippets for bulk get/delete function

rsgowman:
Corresponding doc change to enable it: http://cl/282953279

### DIFF
--- a/snippets/auth.go
+++ b/snippets/auth.go
@@ -185,6 +185,35 @@ func getUserByPhone(ctx context.Context, client *auth.Client) *auth.UserRecord {
 	return u
 }
 
+func bulkGetUsers(ctx context.Context, client *auth.Client) {
+	uid := "uid1"
+	email := "user@example.com"
+	phone := "+15555551234"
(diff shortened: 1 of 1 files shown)


snippets/auth.go
```diff
@@ -185,6 +185,35 @@ func getUserByPhone(ctx context.Context, client *auth.Client) *auth.UserRecord {
...
 
+func bulkGetUsers(ctx context.Context, client *auth.Client) {
+	uid := "uid1"
+	email := "user@example.com"
+	phone := "+15555551234"
+	providerID := "google.com"
+	providerUID := "google_uid1"
+	// [START bulk_get_users_golang]
+	getUsersResult, err := client.GetUsers(ctx, []auth.UserIdentifier{
+		auth.UIDIdentifier{UID: uid},
```
hiranya911, line 196: May be inline the values of uid, email etc for clarity.

rsgowman: Done. I've also applied this to bulk delete (despite it just being uids.)

PR #331, opened 2020-02-11
chore: Added Actions-based release workflow

hiranya911:
We need to take a slightly different approach with Go releases since there are no artifacts to speak of. The proposed release process is:

1. Create a PR against master and apply the label `release:stage`. This will trigger the staging workflow and run the tests. (Actually the label is optional in this case. Simply targeting the master branch is sufficient)
2. Merge the PR once all the reviews and checks are completed. Code is now released.
3. The publish action triggers at this point which creates the tag.

Sample staging run: https://github.com/firebase/firebase-admin-go/runs/439827992


.github/workflows/publish.yml
```diff
@@ -0,0 +1,64 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Publish Release
+
+on:
+  # Only run the workflow when a PR is merged to the master branch.
+  pull_request:
+    branches: master
+    types: closed
```
lahirumaramba: If we close a PR without merging would that also trigger this?

hiranya911: It does, but then will get immediately skipped because of:

```
if: github.event.pull_request.merged
```

Issue #332, opened 2020-02-12
FR Add error code in messaging.ErrorInfo


p1ass:
### Describe your environment
  * Library version: v3.12.0
  * Firebase Product: Messaging 

### Describe the problem

I would like to switch errors handling according to the response of `SubscribeToTopic()`. 
However, `messaging.ErrorInfo` which is part of the response has only string reason, but not error code. So it is difficult to handle errors.
I think `messaging.ErrorInfo` should have an error code to make error handling more easier.

Since the error code has already used inside the library, it is not difficult to change it to be public, I think.
https://github.com/firebase/firebase-admin-go/blob/master/messaging/messaging.go#L58-L106  

#### Relevant Code:

The suggested change is below:

```go
type ErrorInfo struct {
	Index  int
	Reason string
    Code string // Added
}
```

https://github.com/firebase/firebase-admin-go/blob/master/messaging/messaging.go#L895-L899

hiranya911:
Sounds like a reasonable addition to me.

p1ass:
@hiranya911 Can I send a pull request about this issue?

hiranya911:
@p1ass yes, that would be great!

hiranya911:
This has been fixed for a while (see #343), and released in v4

PR #333, opened 2020-02-16
feat(fcm): Add error code in messaging.ErrorInfo

p1ass:
## Issue

#332 

## What
- Add error code field in `messaging.ErrorInfo`.


p1ass:
Do I have to change more code?

hiranya911:
This has been addressed in the v4 branch via #343. There, the backend error code is directly exposed as the `Reason` field. This will be released within a couple of months. In the meantime it's best to not introduce new additions to the API, especially since those additions will become breaking API changes when v4 comes out.

p1ass:
Oh, I missed #343 pull request! Thank you.

PR #335, opened 2020-02-20
fix(fcm): correct the iidEndpoint endpoints used for topic management

rueian:
According to the document https://developers.google.com/instance-id/reference/server,
the endpoints should be:

https://iid.googleapis.com/iid/v1:batchAdd
https://iid.googleapis.com/iid/v1:batchRemove

NOT:

https://iid.googleapis.com/iid/v1/:batchAdd
https://iid.googleapis.com/iid/v1/:batchRemove

RELEASE NOTE: Updated the remote endpoint used by the topic management operations.

### DIFF
--- a/messaging/topic_mgt.go
+++ b/messaging/topic_mgt.go
@@ -26,8 +26,8 @@ import (
 
 const (
 	iidEndpoint    = "https://iid.googleapis.com/iid/v1"
-	iidSubscribe   = ":batchAdd"
-	iidUnsubscribe = ":batchRemove"
+	iidSubscribe   = "batchAdd"
+	iidUnsubscribe = "batchRemove"
 )
 
 var iidErrorCodes = map[string]struct{ Code, Msg string }{
@@ -164,7 +164,7 @@ func (c *iidClient) makeTopicManagementRequest(ctx context.Context, req *iidRequ
 
 	request := &internal.Request{
 		Method: http.MethodPost,
-		URL:    fmt.Sprintf("%s/%s", c.iidEndpoint, req.op),
+		URL:    fmt.Sprintf("%s:%s", c.iidEndpoint, req.op),
 		Body:   internal.NewJSONEntity(req),
 	}
 	var result iidResponse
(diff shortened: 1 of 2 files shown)


messaging/topic_mgt.go
```diff
@@ -25,9 +25,9 @@ import (
 )
 
 const (
-	iidEndpoint    = "https://iid.googleapis.com/iid/v1"
-	iidSubscribe   = ":batchAdd"
-	iidUnsubscribe = ":batchRemove"
+	iidEndpoint    = "https://iid.googleapis.com/iid"
```
hiranya911: Instead of this, let's change how the URL is constructed in the `makeTopicManagementRequest()` function.

```
iidSubscribe = "batchAdd"
iidUnsubscribe = "batchRemove"

fmt.Sprintf("%s:%s", c.iidEndpoint, req.op)
```

This will cause a test failure, so we need to update the tests too.

rueian:
Thank you for reviewing my PR.

I have changed the base branch to dev, and fix the tests and the `makeTopicManagementRequest`.

I am not sure that both `/v1/:suffix` and `/v1:suffix` are resolved to the same endpoint:
When using invalid or wrong access token to request `/v1/:suffix`, it will response a html page:
```
HTTP/1.1 401 Unauthorized
Content-Type: text/html; charset=UTF-8
Accept-Ranges: none
Vary: Accept-Encoding
Connection: close

<HTML>
<HEAD>
<TITLE>Unauthorized</TITLE>
</HEAD>
<BODY BGCOLOR="#FFFFFF" TEXT="#000000">
<H1>Unauthorized</H1>
<H2>Error 401</H2>
</BODY>
</HTML>
```

However when requesting to `/v1:suffix`, it will response a json:
```
HTTP/1.1 401 Unauthorized
Content-Type: application/json; charset=UTF-8
Connection: close

{"error":"Not authenticated or unauthorized"}
```

Another question, is there any plan to support the `https://iid.googleapis.com/iid/v1:batchImport`?

PR #340, opened 2020-03-02
Fix tenant custom token

xyzdevops:
I had a stab at it.

Please can someone approve or finish the job so Go can use the full features.


hiranya911:
This was addressed in #371

PR #344, opened 2020-03-05
feat(auth): Add ability to link a federated id with the UpdateUser() method.

rsgowman:


rsgowman:
Note the diffbase. I'll merge+retarget to dev once the bulk_get stuff lands.

auth/user_mgt_test.go
```diff
@@ -751,6 +793,37 @@ var updateUserCases = []struct {
...
+			UID:        "google_uid",
+		}),
+		map[string]interface{}{
+			"linkProviderUserInfo": &UserProvider{
+				ProviderID: "google.com",
+				UID:        "google_uid",
+			}},
+	},
+	{
+		(&UserToUpdate{}).PhoneNumber("").ProvidersToDelete([]string{"google.com"}),
```
hiranya911: Also test for email and phone

integration/auth/user_mgt_test.go
```diff
@@ -368,37 +368,221 @@ func TestCreateUser(t *testing.T) {
...
+		userRecord, err := client.UpdateUser(context.Background(), updateUser.UID, params)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Ensure link operation worked as expected
+		providerUIDs := mapToProviderUIDs(userRecord.ProviderUserInfo)
+		providerIDs := mapToProviderIDs(userRecord.ProviderUserInfo)
+		if !contains(providerUIDs, googleFederatedUID) {
+			t.Errorf("UpdateUser().ProviderUserInfo[*].UID want include %v; got %v",
```
hiranya911: Typically assertion failures are reported as follows.

```
t.Errorf("UpdateUser().ProviderUserInfo[*].UID = %v; want incldue = %q", providerUIDs, googleFederatedUID)
```

integration/auth/user_mgt_test.go
```diff
@@ -368,37 +368,221 @@ func TestCreateUser(t *testing.T) {
...
+				googleFederatedUID, providerUIDs)
+		}
+		if !contains(providerIDs, "google.com") {
+			t.Errorf("UpdateUser().ProviderUserInfo[*].ProviderID want include 'google.com'; got %v",
+				providerIDs)
+		}
+
+		// Unlink user from federated provider
```
hiranya911: Perhaps move this part into a separate `t.Run()` context?

integration/auth/user_mgt_test.go
```diff
@@ -368,37 +368,221 @@ func TestCreateUser(t *testing.T) {
...
+			t.Errorf("UpdateUser().ProviderUserInfo[*].ProviderID want not include all %v; got %v",
+				notWantAll, providerIDs)
+		}
+	})
+
+	t.Run("NoopsGivenEmptyProvidersToDelete", func(t *testing.T) {
+		userRecord := createTestUser("NoopWithEmptyProvidersToDeleteUser")
+		defer deleteUser(userRecord.UID)
+
+		gotUserRecord, err := client.UpdateUser(context.Background(), userRecord.UID,
```
hiranya911: Shouldn't this fail with a validation error. We got:

```
        if len(u.params) == 0 {
		// update without any parameters is never allowed
		return nil, fmt.Errorf("update parameters must not be nil or empty")
	}
```

Perhaps we should change `ProvidersToDelete()` function ignore empty array, if the existing value for that key is already empty?

integration/auth/user_mgt_test.go
```diff
@@ -954,3 +1146,23 @@ func importUser(t *testing.T, uid string, userToImport *auth.UserToImport) *auth
 	}
 	return userRecord
 }
+
+// Helper function that deletes the user with the specified phone number if it
+// exists.
+// TODO(rsgowman): This function was ported from node.js port; a number of tests
+// there use this, but haven't been ported to go yet. Do so.
+func deletePhoneNumberUser(t *testing.T, phoneNumber string) {
```
hiranya911: We really ought to stop passing `t` to helper functions. Instead return an error from this function, and handle it at the callsite.

integration/auth/user_mgt_test.go
```diff
@@ -954,3 +1146,23 @@ func importUser(t *testing.T, uid string, userToImport *auth.UserToImport) *auth
...
+	userRecord, err := client.GetUserByPhoneNumber(context.Background(), phoneNumber)
+	if err != nil {
+		if strings.HasPrefix(err.Error(), "cannot find user from phone number:") {
```
hiranya911: `if auth.IsUserNotFound(err)`

integration/auth/user_mgt_test.go
rsgowman: This pattern comes direct from go/go-style/best-practices#test-helper-error-handling.

I like it, since it reduces clutter at the call site for what's essentially a failed precondition. But I can change it to return an error if you feel strongly.

integration/auth/user_mgt_test.go
rsgowman: Done.

.github/scripts/publish_post_check.sh
```diff
@@ -0,0 +1,105 @@
+
```
hiranya911, line 1: Please remove any changes to Actions workflow configs.

integration/auth/user_mgt_test.go
hiranya911: Oh, that's good to know. I guess I'm always a bit weirded out by code that passes `t` around. But I suppose that's considered normal in Go.

integration/auth/user_mgt_test.go
```diff
@@ -983,3 +1193,23 @@ func importUser(t *testing.T, uid string, userToImport *auth.UserToImport) *auth
...
+	userRecord, err := client.GetUserByPhoneNumber(context.Background(), phoneNumber)
+	if err != nil {
+		if auth.IsUserNotFound(err) {
+			// User already doesn't exist.
+			return
+		}
+		t.Fatal(err)
+	}
+
+	err = client.DeleteUser(context.Background(), userRecord.UID)
```
hiranya911, line 1211: Error check can be inlined here.

hiranya911 (changes requested):
Implementation looks pretty solid. Lets get rid of the CI/workflow stuff that's unrelated.

.github/scripts/publish_post_check.sh
rsgowman: Done. (Must've been a bad merge)

integration/auth/user_mgt_test.go
rsgowman: Done.

hiranya911 (approved):
LGTM 👍

auth/user_mgt.go
```diff
@@ -296,6 +325,78 @@ func (u *UserToUpdate) validatedRequest() (map[string]interface{}, error) {
...
 		}
 	}
+
+	if linkProviderUserInfo, ok := req["linkProviderUserInfo"]; ok {
+		userProvider := linkProviderUserInfo.(*UserProvider)
+		if err := validateProviderUserInfo(userProvider); err != nil {
+			return nil, err
+		}
+
+		// Although we don't really advertise it, we want to also handle linking of
```
egilmorez, line 335: Assuming that none of this comment gets parsed into the generated docs.

rsgowman: No, this shouldn't make it into the public facing docs.

PR #345, opened 2020-03-06
RTDB error handling revamp

hiranya911:
* Migrated the RTDB implementation off the deprecated APIs in `internal`.
* Implement platform error handling for RTDB API calls.

### DIFF
--- a/db/query_test.go
+++ b/db/query_test.go
@@ -16,8 +16,11 @@ package db
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"reflect"
 	"testing"
+
+	"firebase.google.com/go/v4/errorutils"
 )
 
 var sortableKeysResp = map[string]interface{}{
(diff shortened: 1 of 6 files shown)


db/query_test.go
```diff
@@ -768,10 +771,15 @@ func TestQueryHttpError(t *testing.T) {
 
 	want := "http error status: 500; reason: test error"
 	result, err := testref.OrderByChild("child").GetOrdered(context.Background())
-	if err == nil || err.Error() != want {
-		t.Errorf("GetOrdered() = %v; want = %v", err, want)
+	if result != nil || err == nil || err.Error() != want {
+		t.Fatalf("GetOrdered() = (%v, %v); want = (nil, %v)", result, err, want)
```
lahirumaramba: Just curious, why use `Fatal` here instead `Error`?

hiranya911: Rest of the test should not run if this assertion fails.

PR #352, opened 2020-03-18
Expand the scope of accepted valid email addresses

### LINKED ISSUE #351: Email Validation func is Improper
The email validation func checks if the split over `@` results in 2 parts:

https://github.com/firebase/firebase-admin-go/blob/32728b0fc9f9486276937d8f50cdf75485c1293c/auth/user_mgt.go#L461-L463

But this risks rejecting valid email address. These are all valid email addresses:

- Abc\@def@example.com
- "Abc@def"@example.com
- "<\"@\\".!.#%$@example.com

which the library will reject. The recommendation per Stavros Korokithakis is to check if there's an @ symbol and try to send an email to it. If they click the validation link, then it's a real address. So replace the prior snippet with:

```go
if !strings.Contains(email, "@") {
	return fmt.Errorf("malformed email string: %q", email)
}
```

References:
- [I Knew How To Validate An Email Address Until I Read The RFC](https://haacked.com/archive/2007/08/21/i-knew-how-to-validate-an-email-address-until-i.aspx/)
- [So you think you can validate email addresses](https://archive.fosdem.org/2018/schedule/event/email_address_quiz/)

Mohammed90:
It is possible for an email address to contain more than one @ symbol. Validating email addresses is hard and complicated. The best way to validate an email address is to check for the presence of an @, in proper place or not, and send an email to it. Nothing more.

Closes #351.

I edited this on the web interface, and the patch branch is off of master. Will this be an issue, even though the PR is onto dev?

### DIFF
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -458,7 +458,7 @@ func validateEmail(email string) error {
 	if email == "" {
 		return fmt.Errorf("email must be a non-empty string")
 	}
-	if parts := strings.Split(email, "@"); len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+	if !strings.Contains(email, "@") {
 		return fmt.Errorf("malformed email string: %q", email)
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


Mohammed90:
I'll do it properly off the `dev` branch

PR #353, opened 2020-03-18
Handling timeouts, connection and other network errors

hiranya911:
Handling low-level transport errors:
  * Timeouts: DeadlineExceeded
  * Connection errors: Unavailable
  * Everything else: Unknown

I'm using `os.IsTimeout()` function to detect timeout errors. There's no similar util for detecting connection errors, so I implemented one. Most errors returned by the `http` package are wrapped in `url.Error`, so we need to unwrap it to get to the original error. Some of the error conditions this implementation looks for is Linux specific, but probably good enough for us to start with.

### DIFF
--- a/internal/errors.go
+++ b/internal/errors.go
@@ -17,7 +17,11 @@ package internal
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/http"
+	"net/url"
+	"os"
+	"syscall"
 )
 
 // ErrorCode represents the platform-wide error codes that can be raised by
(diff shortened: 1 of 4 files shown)


internal/errors.go
```diff
@@ -148,3 +152,46 @@ func NewFirebaseErrorOnePlatform(resp *Response) *FirebaseError {
...
+	}
+
+	return &FirebaseError{
+		ErrorCode: code,
+		String:    msg,
+		Ext:       make(map[string]interface{}),
+	}
+}
+
+// isConnectionRefused attempts to determine if the given was caused by a failure to establish a
```
lahirumaramba:
```suggestion
// isConnectionRefused attempts to determine if the given error was caused by a failure to establish a
```

internal/errors.go
```diff
@@ -148,3 +152,46 @@ func NewFirebaseErrorOnePlatform(resp *Response) *FirebaseError {
...
+// connection.
+//
+// A net.OpError where the Op field is set to "dial" or "read" is considered a connection refused
+// error. Similarly an ECONNREFUSED error code (Linux-specific) is also considered a connection
+// refused error.
+func isConnectionRefused(err error) bool {
+	switch t := err.(type) {
+	case *url.Error:
+		return isConnectionRefused(t.Err)
```
lahirumaramba: Do we usually need to unwrap all errors twice? Just thinking out loud to see any potential complications having a recursive call here...

hiranya911: `Err` is a member of `url.Error` and `net.OpError` types. So we cannot really unwrap other error types.

Issue #99, opened 2018-02-18
FR : list of topic subscriptions


chemidy:
We can now SubscribeToTopic and UnsubscribeFromTopic.

Is it possible to add a function ListTopicsForDevice based on "rel/topics" attribute  [from the Instance ID service](https://developers.google.com/instance-id/reference/server#get_information_about_app_instances) ?

hiranya911:
Sounds like a useful feature to me. @sampson-chen any thoughts?

chemidy:
Any update ? Can I send a PR ?

hiranya911:
@chemidy Before putting any time into writing code, can you propose what this API is going to look like (method/type signatures)? I can take that to the FCM team and get their thoughts on it.

chong-shao:
Hi, we have plan to add this feature to the AdminSDK (along with a new HTTP API). But we don't have an estimated timeline now.

PR #362, opened 2020-04-16
Added topic subscription information retrieval to messaging client

m4ng0:
This commit should provide the feature requested in FR #99.

### DIFF
--- a/messaging/topic_mgt.go
+++ b/messaging/topic_mgt.go
@@ -20,14 +20,17 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"firebase.google.com/go/internal"
 )
 
 const (
-	iidEndpoint    = "https://iid.googleapis.com/iid/v1"
-	iidSubscribe   = "batchAdd"
-	iidUnsubscribe = "batchRemove"
+	iidBaseEndpoint = "https://iid.googleapis.com/iid"
+	iidEndpoint     = iidBaseEndpoint + "/v1"
+	iidInfoEndpoint = iidBaseEndpoint + "/info"
+	iidSubscribe    = "batchAdd"
+	iidUnsubscribe  = "batchRemove"
 )
 
 var iidErrorCodes = map[string]struct{ Code, Msg string }{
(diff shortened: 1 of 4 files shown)


hiranya911:
@chong-shao is this something the FCM team is interested in releasing?

messaging/topic_mgt.go
```diff
@@ -20,14 +20,17 @@ import (
...
+	"time"
 
 	"firebase.google.com/go/internal"
 )
 
 const (
-	iidEndpoint    = "https://iid.googleapis.com/iid/v1"
-	iidSubscribe   = "batchAdd"
-	iidUnsubscribe = "batchRemove"
+	iidBaseEndpoint = "https://iid.googleapis.com/iid"
```
andirayo, line 29: Note:
We are in the process of deprecating everything related to Firebase Instance ID in the year 2020.
All Topic related features will move into Firebase Messaging and most likely the endpoint(s)s will move to be an FCM domain.
Thus, I recommend not to add new client logic that depends on a server endpoint that we plan to deprecate soon.

Maybe, we can the same endpoints on the FCM domain `fcm.googleapis.com` and use those URLs instead.
Probably we can just add a URL `/topics` to the existing GFE config file of `fcm.googleapis.com` and point it to the same IID logic (server-side).
That way we could release new client software with that URL and without using endpoints that are about to be deprecated.

chong-shao: Follow up on andirayo@'s comment: 
The timeline of moving to endpoints under fcm.googleapis.com is not determined. We decide to not block this PR. 
Please feel free to use iid.googleapis.com in the PR.

hiranya911:
@chong-shao there's a product decision to be made here. Do we even want to expose this API in our SDKs? We ought to run this by a PM first.

chong-shao:
@hiranya911 synced with our PM. The feedback is that we should not merge this into master branch unless the endpoint has been switched to the new one (which has not been set up yet).

Can we put this is a side branch? Or keeping the PR in open status until the new endpoints are ready? @andirayo

hiranya911:
@chong-shao I agree. We can keep the PR open for now.

We also possibly need an API proposal, and at least a semblance of a plan to have the same API exposed in other languages at some point.

messaging/topic_mgt.go
andirayo: +1

PR #373, opened 2020-05-15
fix(fcm): Add ability to override default FCM endpoint via ClientOptions

### LINKED ISSUE #189: Unable to set endpoint for messaging
I try to set the base url for the client library so that during testing I can run requests against a FCM mock instead of doing internet requests. However, even with the following code, I can see the attempts to reach the official base url:

```go
ctx := context.Background()
opts := []option.ClientOption{option.WithEndpoint("http://localhost:1234/")}
// opts = append(opts, option.WithoutAuthentication())
app, _ := firebase.NewApp(ctx, nil, opts...)
c, _ := firebaseApp.Messaging(ctx)

_, _ = c.Send(cox, nil)
```

```
POST /v1/projects/dummy-project-id/messages:send HTTP/1.1
Host: fcm.googleapis.com
User-Agent: Go-http-client/1.1
Content-Length: 71
Content-Type: application/json
Accept-Encoding: gzip
{"message":{"notification":{"title":"foo","body":"bar"},"token":"123"}}
```

```
HTTP/2.0 401 Unauthorized
Alt-Svc: quic=":443"; ma=2592000; v="44,43,39,35"
Cache-Control: private
Content-Type: application/json; charset=UTF-8
Date: Mon, 05 Nov 2018 18:37:10 GMT
Server: ESF
Vary: Origin
Vary: X-Origin
Vary: Referer
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 1; mode=block

{
  "error": {
    "code": 401,
    "message": "Request is missing required authentication credential. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.",
    "status": "UNAUTHENTICATED"
  }
}
```

slirx:
Fixes #189


slirx:
I've updated it to be able to use the endpoint set via `options.WithEndpoint()`.
I've used `internaloption.WithDefaultEndpoint()` in order to apply the default endpoint.

messaging/messaging.go
```diff
@@ -913,13 +914,14 @@ func NewClient(ctx context.Context, c *internal.MessagingConfig) (*Client, error
 		return nil, errors.New("project ID is required to access Firebase Cloud Messaging client")
 	}
 
-	hc, _, err := transport.NewHTTPClient(ctx, c.Opts...)
+	c.Opts = append(c.Opts, internaloption.WithDefaultEndpoint(messagingEndpoint))
```
hiranya911, line 917: Let's implement this without bringing in yet another package.

```
hc, endpoint, err := transport.NewHTTPClient(ctx, c.Opts...)
if err != nil {
		return nil, err
}

if endpoint == "" {
  endpoint = messagingEndpoint
}
```

slirx: done :) also I've added tests

messaging/messaging_test.go
```diff
@@ -1126,6 +1126,43 @@ func TestSend(t *testing.T) {
...
+		w.Header().Set("Content-Type", "application/json")
+		w.Write([]byte("{ \"name\":\"" + testMessageID + "\" }"))
+	}))
+	defer ts.Close()
+
+	ctx := context.Background()
+
+	conf := *testMessagingConfig
+	optEndpoint := option.WithEndpoint(ts.URL)
+	conf.Opts = append(conf.Opts, optEndpoint)
```
hiranya911, line 1144: No need to append here. Only `WithEndpoint()` should be enough:

```
conf.Opts = []option.ClientOption{
    option.WithEndpoint(ts.URL),
}
```

firebase_test.go
```diff
@@ -361,6 +362,44 @@ func TestMessaging(t *testing.T) {
...
+
+	ctx := context.Background()
+
+	tokenSource := &testTokenSource{AccessToken: "mock-token-from-custom"}
+	app, err := NewApp(
+		ctx,
+		nil,
+		option.WithCredentialsFile("testdata/service_account.json"),
+		option.WithTokenSource(tokenSource),
+		option.WithEndpoint(ts.URL),
```
hiranya911, line 382: Only `WithEndpoint()` should be sufficient here.

firebase_test.go
```diff
@@ -361,6 +362,44 @@ func TestMessaging(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	c, err := app.Messaging(ctx)
+	if c == nil || err != nil {
+		t.Fatalf("Messaging() = (%v, %v); want (iid, nil)", c, err)
+	}
+
+	msg := &messaging.Message{
+		Android: nil,
```
hiranya911, line 394: Remove redundant.

messaging/messaging_test.go
slirx: It gives an error without append:
```go
conf := *testMessagingConfig
conf.Opts = []option.ClientOption{
    option.WithEndpoint(ts.URL),
}

client, err := NewClient(ctx, &conf)
if err != nil {
    t.Fatal(err)
}
```
```
--- FAIL: TestSendWithCustomEndpoint (0.34s)
    messaging_test.go:1149: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.
```

firebase_test.go
slirx: It gives an error only with `WithEndpoint()`:
```go
app, err := NewApp(ctx, nil, option.WithEndpoint(ts.URL))
if err != nil {
    t.Fatal(err)
}

c, err := app.Messaging(ctx)
if c == nil || err != nil {
    t.Fatalf("Messaging() = (%v, %v); want (iid, nil)", c, err)
}
```
```
--- FAIL: TestMessagingSendWithCustomEndpoint (0.59s)
    firebase_test.go:390: Messaging() = (<nil>, project ID is required to access Firebase Cloud Messaging client); want (iid, nil)
```

firebase_test.go
slirx: done. I've removed `Android: nil,`.

firebase_test.go
hiranya911: Ok, I suppose the transport still requires some sort of a credential. In that case let's keep `option.WithTokenSource()`. You definitely don't need `WithCredentialsFile()` when a TokenSource is explicitly provided.

messaging/messaging_test.go
hiranya911: Looks like the transport is still looking for credentials, which is ok. We can leave this as is.

firebase_test.go
slirx: It seems `WithCredentialsFile()` is required to provide project id. It gives an error without `WithCredentialsFile()`:

```go
tokenSource := &testTokenSource{AccessToken: "mock-token-from-custom"}
app, err := NewApp(
    ctx,
    nil,
    //option.WithCredentialsFile("testdata/service_account.json"),
    option.WithTokenSource(tokenSource),
    option.WithEndpoint(ts.URL),
)
if err != nil {
    t.Fatal(err)
}

c, err := app.Messaging(ctx)
if c == nil || err != nil {
    t.Fatalf("Messaging() = (%v, %v); want (iid, nil)", c, err)
}
```

```
--- FAIL: TestMessagingSendWithCustomEndpoint (0.00s)
    firebase_test.go:390: Messaging() = (<nil>, project ID is required to access Firebase Cloud Messaging client); want (iid, nil)
```

hiranya911: Ah! :(

PR #380, opened 2020-06-12
fix: Removed import path comments

hiranya911:
Also bumped the minimum Go version to 1.12.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ requests, code review feedback, and also pull requests.
 
 ## Supported Go Versions
 
-We support Go v1.11 and higher.
+We support Go v1.12 and higher.
 [Continuous integration](https://github.com/firebase/firebase-admin-go/actions) system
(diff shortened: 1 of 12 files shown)


hiranya911:
After a bit of experimentation it seems the safest option is to remove the import paths altogether.

**1. Import paths are not used when used as a module.**

From https://golang.org/cmd/go/#hdr-Remote_import_paths

> Import path checking is also disabled when using modules. Import path comments are obsoleted by the go.mod file's module statement.

**2. Having import paths set to the versioned (`/v4`) package names break compilation when used in the non-module mode.**

```
package main

import (
        "context"
	firebase "firebase.google.com/go"
)

func main() {
	app, err := firebase.NewApp(context.Background(), nil)
	// ...
}
```

```
main.go:8:2: code in directory gopath/src/firebase.google.com/go expects import "firebase.google.com/go/v4"
```

The import path is basically making it impossible for non-modules users to import the code.

PR #384, opened 2020-06-16
[chore] Release 4.0.0 - take 2

hiranya911:
Fixed a flaky auth integration test by retrying the `GetUser()` API call a few times with a small delay.

### DIFF
--- a/integration/auth/user_mgt_test.go
+++ b/integration/auth/user_mgt_test.go
@@ -249,9 +249,22 @@ func TestLastRefreshTime(t *testing.T) {
 		t.Errorf("signInWithPassword failed: %v", err)
 	}
 
-	getUsersResult, err := client.GetUser(context.Background(), userRecord.UID)
-	if err != nil {
-		t.Fatalf("GetUser(...) failed with error: %v", err)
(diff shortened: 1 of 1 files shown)


integration/auth/user_mgt_test.go
```diff
@@ -249,9 +249,22 @@ func TestLastRefreshTime(t *testing.T) {
...
-	getUsersResult, err := client.GetUser(context.Background(), userRecord.UID)
-	if err != nil {
-		t.Fatalf("GetUser(...) failed with error: %v", err)
+	// Attempt to retrieve the user 3 times (with a small delay between each attempt.) Occasionally,
+	// this call retrieves the user data without the lastLoginTime/lastRefreshTime fields; possibly
+	// because it's hitting a different server than what the login request used.
+	var getUsersResult *auth.UserRecord
+	for i := 0; i < 3; i++ {
+		var err error
+		getUsersResult, err = client.GetUser(context.Background(), userRecord.UID)
```
rsgowman, line 258: nit: Could you use `:=` here and eliminate the var err statement on the previous line?

Optional.

hiranya911: That also declares a new `getUsersResult` variable in the nested scope. We want to reuse the one defined outside the for loop.

PR #392, opened 2020-08-04
feat(auth): Hash Order Fix

Galadros:
Allow users to specify inputHashOrder (salt first or password first) for hmac{md5, sha1, sha256, sha512} and {sha1, sha256, sha512}, persuant to an internal API proposal (similar features are being introduced for each Firebase Admin language).   Also updated CONTRIBUTING.md to explain why some tests will fail for a non-GCIP Firebase project.

RELEASE NOTE: Implemented support for specifying password-salt order in hash configurations used when importing users.


hiranya911:
Is there a Go API proposal for the changes made here? I feel like instead of introducing a new `InputOrder` type and adding new fields to all hash types, we should just introduce 2 new functions:

```
func WithPasswordFirstHash(h UserImportHash) UserImportOption
func WithSaltFirstHash(h UserImportHash) UserImportOption
```

Usage example:
```
h := &hash.SHA512{
  Rounds: 1024,
}
result, err := client.ImportUsers(ctx, users, auth.WithPasswordFirstHash(h))
if err != nil {
  log.Fatalln(err)
}
```

This way the only new additions are the 2 helper functions, and we can handle the input order parameter in just one place instead of many.

integration/auth/auth_test.go
```diff
@@ -260,6 +267,198 @@ func signInWithPassword(email, password string) (string, error) {
...
+		user := (&auth.UserToImport{}).
+			UID(uid).
+			Email(email).
+			PasswordHash(test.localHash.Sum(nil)).
+			PasswordSalt([]byte(salt))
+		result, err := client.ImportUsers(context.Background(), []*auth.UserToImport{user}, auth.WithHash(test.hashConfig))
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer deleteUser(uid)
```
hiranya911, line 438: Have you tried running this test? I'm wondering how it can possibly work. Because `deleteUser()` calls are deferred until all cases are run, it will attempt to import the same `uid` repeatedly, which should fail.

integration/auth/auth_test.go
```diff
@@ -260,6 +267,198 @@ func signInWithPassword(email, password string) (string, error) {
 	return respBody.IDToken, err
 }
 
+func TestSaltPasswordOrder(t *testing.T) {
```
hiranya911, line 270: Let's verify this behavior in the unit tests. That's faster and also gets checked more often in our CI. In the integration tests we can check one hash configuration with input order for completeness. Also add it in the `uset_mgt_test.go` file closer to the other `ImportUsers()` tests.

auth/hash/hash_test.go
```diff
@@ -268,19 +268,19 @@ var invalidHashes = []struct {
...
 	},
 	{
 		name: "SHA256: rounds too high",
-		alg:  SHA256{8193},
+		alg:  SHA256{8193, InputOrderUnspecified},
 	},
 	{
 		name: "SHA512: rounds too high",
-		alg:  SHA512{8193},
+		alg:  SHA512{8193, InputOrderUnspecified},
```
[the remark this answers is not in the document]
Galadros: Do you mean creating a struct with some InputOrder assigned and then checking that that InputOrder is indeed the correct one?  I'm not quite sure what you mean here.

integration/auth/auth_test.go
Galadros: The purpose of all of this is to make sure that the remote server is properly processing the options I gave it, being responsive to whether the password or hash should come first.  Isn't that the definition of an integration test?

hiranya911: Testing if the remote server works correctly is out of scope for us. Our testing should cover the correct serialization of the arguments and the HTTP request/response handling, which can be done easily in the unit tests.

auth/hash/hash_test.go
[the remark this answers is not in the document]
hiranya911: Yes. Make sure the InputOrder argument is correctly serialized for a range of different hash configurations. Similar to how the existing `TestValidHash` test case works. As of now the new functionality has zero unit test coverage.

integration/auth/auth_test.go
Galadros: Fair point, sorry- this is an artifact from a previous testing structure before I moved to this iterative version.  Fixed.

auth/hash/hash_test.go
```diff
@@ -311,3 +311,194 @@ func TestInvalidHash(t *testing.T) {
...
+	{
+		alg: SHA512{Rounds: 8192},
+		want: internal.HashConfig{
+			"hashAlgorithm": "SHA512",
+			"rounds":        8192,
+		},
+	},
+}
+
+var inputOrders = []InputOrderType{
```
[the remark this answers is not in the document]
Galadros: Entirely correct, sorry- this is a relic from a previous setup that I forgot to remove.

integration/auth/auth_test.go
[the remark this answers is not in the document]
Galadros: Sorry, I thought that I'd cleaned up all but two of these.  I was told to implement integration testing for this change, so I'm leaving two in here to make sure that the remote end of things is playing nice.

integration/auth/auth_test.go
```diff
@@ -260,6 +265,85 @@ func signInWithPassword(email, password string) (string, error) {
...
+	uid := randomUID()
+	email := randomEmail(uid)
+	password := "pass123123"
```
hiranya911, line 271: Use a `const ()` block here.

integration/auth/auth_test.go
```diff
@@ -260,6 +265,85 @@ func signInWithPassword(email, password string) (string, error) {
...
+		{
+			testName: "HMAC_SHA256_PasswordFirst",
+			hashConfig: hash.HMACSHA256{
+				Key:        []byte(key),
+				InputOrder: hash.InputOrderPasswordFirst,
+			},
+			localHash: hmac.New(sha256.New, []byte(key)),
+		},
+	}
+	// h := hash.HMACSHA256{
```
hiranya911, line 296: Please remove

integration/auth/auth_test.go
```diff
@@ -260,6 +265,85 @@ func signInWithPassword(email, password string) (string, error) {
...
+	// 	Key: []byte(key),
+	// 	InputOrder: InputOrderSaltFirst
+	// }
+	// myHash := hmac.New(sha256.New, []byte(key))
+	// myHash.Write([]byte(salt+password))
+	for _, test := range tests {
+		hC, _ := test.hashConfig.Config()
```
hiranya911, line 303: We should avoid using internal APIs in integration tests. You can include a bool or an InputOrderType value in the test config struct instead.

```
if (test.passwordFirst) {

} else {
  
}
```

Or (possibly better), turn `test.localHash` into a `func() []byte`:

```
{
  localHash: func() []byte {
    h := sha1.New()
    h.Write([]byte(password + salt)
    return h.Sum(nil)
  }
}

```

integration/auth/auth_test.go
hiranya911: I think this issue still exists. This is using the same `uid` value in multiple iterations of the test. You should either use unique uid+email for each iteration, or not defer the calls to `deleteUser()`.

go.mod
```diff
@@ -5,7 +5,11 @@ go 1.11
 require (
 	cloud.google.com/go/firestore v1.1.1
 	cloud.google.com/go/storage v1.0.0
+	golang.org/x/lint v0.0.0-20200302205851-738671d3881b // indirect
```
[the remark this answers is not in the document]
Galadros: Sounds like a good idea.  For what it's worth, there's nothing atypical about the setup here, so I have no idea what could be causing this.

integration/auth/auth_test.go
Galadros: Do you want me to manually specify the results of hashConfig.Config instead, then?  I thought that I should be using our own functions to test that this actually works (or should I only use those in non-integration tests?)

hiranya911: Yes, internal APIs are ok for unit tests where we already employ a mocks and other tricks to manipulate the API behavior. These integration tests should be implemented more or less from the developer's standpoint without calling any internal APIs.

integration/auth/auth_test.go
```diff
@@ -260,6 +264,78 @@ func signInWithPassword(email, password string) (string, error) {
 	return respBody.IDToken, err
 }
 
+func TestSaltPasswordOrder(t *testing.T) {
```
hiranya911, line 267: Nit: Rename to TestImportUserPasswordSaltOrder

integration/auth/auth_test.go
```diff
@@ -260,6 +264,78 @@ func signInWithPassword(email, password string) (string, error) {
...
+	const (
+		password = "pass123123"
+		key      = "skeleton"
+		salt     = "NaCl"
+	)
+	tests := []struct {
+		testName   string
```
hiranya911, line 274: Just `name`. Also include this in the error logs:

```
t.Errorf("ImportUsers(%s) = %#v; want = {SuccessCount: 1, FailureCount: 0}", test.name, result)
```

Issue #391, opened 2020-07-30
FR: Remote config support


tedgonzalez:
### [REQUIRED] Step 2: Describe your environment
  * Firebase Product: Remote config

### [REQUIRED] Step 3: Describe the problem

Hi do you have plans to support firebase remote config?

lahirumaramba:
Hi Theodore! Thank you for the feature request!
We have recently added Remote Config support in the Admin Node.js SDK. We plan to add the Remote Config API support across the other Admin SDK platforms, including Go. Though I cannot promise you a timeline, this issue will track our progress as we start implementation.

We also greatly appreciate contributions to our codebase, if you would like to contribute to the project by adding Remote Config support! :)

allentv:
I am thinking of proposing Remote Config for a project and would like to have a working version soon. So started working on it. Here is a skeleton based off the work done in the Node.js version : https://github.com/firebase/firebase-admin-go/compare/dev...allentv:allen/add-remote-config?expand=1

lahirumaramba:
Hi @allentv,
That is great to hear! Thank you for your contribution and we really appreciate it!
Let's start reviewing your PR #396 and we can go from there.

allentv:
The current implementation of #396 is a 1:1 mapping of the [REST API docs](https://firebase.google.com/docs/reference/remote-config/rest#service-endpoint) as shown below:

```go
// Remote Config Client
type Client struct {
	hc        *internal.HTTPClient
	projectID string
}

// Factory function for the client
func (a *App) RemoteConfig(ctx context.Context) (*remoteconfig.Client, error)


// API response
type Response struct {
	RemoteConfig
	Etag string `json:"etag"`
}

// Fetch the config
func (c *Client) GetRemoteConfig(versionNumber string) (*Response, error)

// Update the config based on the ETag from the fetch and run optional validation
func (c *Client) UpdateRemoteConfig(eTag string, validateOnly bool) (*Response, error)

// List the versions
func (c *Client) ListVersions(options *ListVersionsOptions) (*ListVersionsResponse, error)

// Rollback a specific version
func (c *Client) Rollback(versionNumber string) (*Template, error) 
```

Since fetching the remote config would yield the complete list of parameter values, I propose to add the following methods:
```go
func (c *Client) Get<Type>(key string) <Type> {}

// Examples:
func (c *Client) GetBool(key string) (bool, error) {}
func (c *Client) GetInt(key string) (int, error) {}
[...]

PR #396, opened 2020-08-08
Add support for Remote Config

allentv:
**Summary**
This PR adds support for Remote Config based on the REST API reference. I ran these methods by invoking them from another file with service account credentials. Tests are pending.

**Related Issue : ** https://github.com/firebase/firebase-admin-go/issues/391

### DIFF
--- a/remoteconfig/client.go
+++ b/remoteconfig/client.go
@@ -1,0 +1,181 @@
+package remoteconfig
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"firebase.google.com/go/v4/internal"
+)
+
+const (
+	defaultRemoteConfigEndpoint = "https://firebaseremoteconfig.googleapis.com/v1"
+)
+
+// Client .
+type Client struct {
+	hc        *internal.HTTPClient
+	projectID string
+}
+
+func (c *Client) getRootURL() string {
+	return fmt.Sprintf("%s/projects/%s/remoteConfig", defaultRemoteConfigEndpoint, c.projectID)
+}
+
+// GetRemoteConfig https://firebase.google.com/docs/reference/remote-config/rest/v1/projects/getRemoteConfig
+func (c *Client) GetRemoteConfig(versionNumber string) (*Response, error) {
+	var opts []internal.HTTPOption
+
+	// Optional. Version number of the RemoteConfig to look up.
+	// If not specified, the latest RemoteConfig will be returned.
+	if versionNumber != "" {
+		opts = append(opts, internal.WithQueryParam("versionNumber", versionNumber))
+	}
+
+	var data RemoteConfig
+
+	resp, err := c.hc.DoAndUnmarshal(
+		context.Background(),
+		&internal.Request{
+			Method: http.MethodGet,
+			URL:    c.getRootURL(),
+			Opts:   opts,
+		},
+		&data,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	result := &Response{
+		RemoteConfig: data,
+		Etag:         resp.Header.Get("Etag"),
+	}
+
+	return result, nil
+}
+
+// UpdateRemoteConfig https://firebase.google.com/docs/reference/remote-config/rest/v1/projects/updateRemoteConfig
+func (c *Client) UpdateRemoteConfig(eTag string, validateOnly bool) (*Response, error) {
+	if eTag == "" {
+		eTag = "*"
+	}
+
+	var opts []internal.HTTPOption
+	opts = append(opts, internal.WithHeader("If-Match", eTag))
+	opts = append(opts, internal.WithQueryParam("validateOnly", strconv.FormatBool(validateOnly)))
+
+	var data RemoteConfig
+
+	resp, err := c.hc.DoAndUnmarshal(
+		context.Background(),
+		&internal.Request{
+			Method: http.MethodPut,
+			URL:    c.getRootURL(),
+			Opts:   opts,
+		},
+		&data,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	result := &Response{
+		RemoteConfig: data,
+		Etag:         resp.Header.Get("Etag"),
+	}
+
+	return result, nil
+}
+
+// ListVersions https://firebase.google.com/docs/reference/remote-config/rest/v1/projects.remoteConfig/listVersions
+func (c *Client) ListVersions(options *ListVersionsOptions) (*ListVersionsResponse, error) {
+	var opts []internal.HTTPOption
+	if options.PageSize != 0 {
+		opts = append(opts, internal.WithQueryParam("pageSize", strconv.FormatInt(options.PageSize, 10)))
+	}
+
+	if options.PageToken != "" {
+		opts = append(opts, internal.WithQueryParam("pageToken", options.PageToken))
+	}
+
+	if options.EndVersionNumber != "" {
+		opts = append(opts, internal.WithQueryParam("endVersionNumber", options.EndVersionNumber))
+	}
+
+	if options.StartTime != "" {
+		opts = append(opts, internal.WithQueryParam("startTime", options.StartTime))
+	}
+
+	if options.EndTime != "" {
+		opts = append(opts, internal.WithQueryParam("endTime", options.EndTime))
+	}
+
+	var data ListVersionsResponse
+	url := fmt.Sprintf("%s:listVersions", c.getRootURL())
+
+	_, err := c.hc.DoAndUnmarshal(
+		context.Background(),
+		&internal.Request{
+			Method: http.MethodGet,
+			URL:    url,
+			Opts:   opts,
+		},
+		&data,
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &data, nil
+}
+
+// Rollback https://firebase.google.com/docs/reference/remote-config/rest/v1/projects.remoteConfig/rollback
+func (c *Client) Rollback(versionNumber string) (*Template, error) {
+	if versionNumber == "" {
+		return nil, errors.New("versionNumber is required to rollback a Remote Config template")
+	}
+
+	var data Template
+	url := fmt.Sprintf("%s:listVersions", c.getRootURL())
+
+	_, err := c.hc.DoAndUnmarshal(
+		context.Background(),
+		&internal.Request{
+			Method: http.MethodPost,
+			Body: internal.NewJSONEntity(
+				struct {
+					VersionNumber string `json:"versionNumber"`
+				}{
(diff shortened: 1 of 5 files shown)


remoteconfig/types.go
```diff
@@ -0,0 +1,90 @@
...
+
+// Response to save the API response including ETag
+type Response struct {
+	RemoteConfig
+	Etag string `json:"etag"`
+}
+
+// Parameter .
+type Parameter struct {
+	ConditionalValues map[string]ParameterValue `json:"conditionalValues"`
```
tedgonzalez, line 58: `ConditionalValues` is optional. How is that handled? Shouldn't you use *map[string]ParameterValue ?

tedgonzalez: Ah got it, it becomes map[string]ParameterValue(nil)

hiranya911:
@lahirumaramba this needs to go through the API review process. I'd rather have an agreed upon API design before diving into the implementation.

allentv:
@hiranya911 Let me know if you require me to fill out any documentation that would be useful for the review process.

hiranya911:
Hi @allenktv. Really appreciate you putting this together. I think it would be helpful if you can add an outline of the proposed public API to #391. Something similar in format to: https://github.com/firebase/firebase-admin-dotnet/issues/149#issuecomment-592145232

Then we can comment on it, and @lahirumaramba  can also start the internal review process based on that.

lahirumaramba:
Hi @allentv !
Please check the proposal I added to #391 
When you get a chance, could you change the base branch of this PR to `remote-config`, please?
That way once the code is updated to match the proposed API, we can merge your changes in. You do not have to implement the complete API. Thanks!

allentv:
I have switched the based branch and made the proposed API spec changes

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+
+	if options.PageToken != "" {
+		opts = append(opts, internal.WithQueryParam("pageToken", options.PageToken))
+	}
+
+	if options.EndVersionNumber != "" {
+		opts = append(opts, internal.WithQueryParam("endVersionNumber", options.EndVersionNumber))
+	}
+
+	if options.StartTime != "" {
```
lahirumaramba, line 108:
```
if !options.StartTime.IsZero() {
....
```

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+		opts = append(opts, internal.WithQueryParam("startTime", options.StartTime))
```
lahirumaramba, line 109:
```
options.StartTime.Format(time.RFC3339Nano)
```

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	if options.EndTime != "" {
```
lahirumaramba, line 112: if !options. EndTime.IsZero() {
....

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+		opts = append(opts, internal.WithQueryParam("endTime", options.EndTime))
```
lahirumaramba, line 113:
```
options. EndTime.Format(time.RFC3339Nano)
```

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// Rollback https://firebase.google.com/docs/reference/remote-config/rest/v1/projects.remoteConfig/rollback
+func (c *Client) Rollback(versionNumber string) (*Template, error) {
+	if versionNumber == "" {
+		return nil, errors.New("versionNumber is required to rollback a Remote Config template")
+	}
+
+	var data Template
+	url := fmt.Sprintf("%s:listVersions", c.getRootURL())
```
lahirumaramba, line 142: I think this should be `rollback`.
```suggestion
	url := fmt.Sprintf("%s:rollback", c.getRootURL())
```

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+	hc        *internal.HTTPClient
+	projectID string
+}
+
+func (c *Client) getRootURL() string {
+	return fmt.Sprintf("%s/projects/%s/remoteConfig", defaultRemoteConfigEndpoint, c.projectID)
+}
+
+// GetRemoteConfig https://firebase.google.com/docs/reference/remote-config/rest/v1/projects/getRemoteConfig
+func (c *Client) GetRemoteConfig(versionNumber string) (*Response, error) {
```
lahirumaramba, line 28: Instead of this function we should have the following as mentioned in the proposed API.
```
func (c *Client) GetTemplate(ctx context.Context) (*Template, error)
func (c *Client) GetTemplateAtVersion(ctx context.Context, versionNumber string) (*Template, error)
```
You could simply change this to `func (c *Client) GetTemplate(ctx context.Context) (*Template, error)` if you don't plan to implement both. We can add `GetTemplateAtVersion()` in a future PR.

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+	result := &Response{
+		RemoteConfig: data,
+		Etag:         resp.Header.Get("Etag"),
+	}
+
+	return result, nil
+}
+
+// UpdateRemoteConfig https://firebase.google.com/docs/reference/remote-config/rest/v1/projects/updateRemoteConfig
+func (c *Client) UpdateRemoteConfig(eTag string, validateOnly bool) (*Response, error) {
```
lahirumaramba, line 61: Same as goes here. In order to stay aligned with the proposed API, this function should be broken down as shown below:
```
func (c *Client) PublishTemplate(ctx context.Context, template *Template) (*Template, error)
func (c *Client) ValidateTemplate(ctx context.Context, template *Template) (*Template, error)
func (c *Client) ForcePublishTemplate(ctx context.Context, template *Template) (*Template, error)
```
Note: they all should return `Template` as well.

If you do not plan to implement all three you can just add `func (c *Client) PublishTemplate(ctx context.Context, template *Template) (*Template, error)` in this PR.

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+	result := &Response{
+		RemoteConfig: data,
+		Etag:         resp.Header.Get("Etag"),
+	}
+
+	return result, nil
+}
+
+// ListVersions https://firebase.google.com/docs/reference/remote-config/rest/v1/projects.remoteConfig/listVersions
+func (c *Client) ListVersions(options *ListVersionsOptions) (*ListVersionsResponse, error) {
```
lahirumaramba, line 94: This should change as:
`func (c *Client) Versions(ctx context.Context, options *ListVersionsOptions) (*VersionIterator, error)`

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	return &Client{
+		hc:        hc,
+		projectID: c.ProjectID,
+	}, nil
+}
+
+// GetTemplate will retrieve the latest template
+func (c *Client) GetTemplate(ctx context.Context) (*Template, error) {
```
lahirumaramba, line 184: Please see my comments above for `GetTemplate` functions. You can implement one if you don't plan to implement other versions.... please remove any placeholder functions.

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// Versions will list the versions of the template
+func (c *Client) Versions(ctx context.Context, options *ListVersionsOptions) (*VersionIterator, error) {
+	// TODO
+	return nil, nil
+}
+
+// Rollback will perform a rollback operation on the template
+func (c *Client) Rollback(ctx context.Context, versionNumber string) (*Template, error) {
```
lahirumaramba, line 202: `Rollback` is already defined above

remoteconfig/client.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// ValidateTemplate will run validations for the current template
+func (c *Client) ValidateTemplate(ctx context.Context, template *Template) (*Template, error) {
+	// TODO
+	return nil, nil
+}
+
+// ForcePublishTemplate will publish the template irrespective of the outcome from validations
+func (c *Client) ForcePublishTemplate(ctx context.Context, template *Template) (*Template, error) {
```
lahirumaramba, line 220: Please remove all the placeholder functions if you don't plan to implement them as part of this PR. Thank you!

remoteconfig/types.go
```diff
@@ -0,0 +1,143 @@
...
+// PageInfo represents the information about a Page
+func (it *VersionIterator) PageInfo() *iterator.PageInfo {
+	// TODO
+	return nil
+}
+
+// Next will return the next version item in the loop
+func (it *VersionIterator) Next() (*Version, error) {
+	return nil, nil
+}
```
lahirumaramba, line 56: These need to be implemented. Feel free to remove the placeholders if you do not plan to add them as part of this PR

lahirumaramba (changes requested):
Thank you @allentv for updating the PR!
I have added a few comments. Please note that we need the public API to match the proposed API in #391 
Please remove all the placeholder functions if you do not plan to include in this PR

lahirumaramba:
@allentv I reconsidered the changes. I think this looks good as it is. Please fix the build errors and we should be good to merge it to the feature branch. That way there will be minimal changes to make. Thanks again for your contribution!

remoteconfig/client.go
lahirumaramba: Reconsidered and this looks like a good first step. It is fine to leave this as it is. No need to make further changes. Please ignore my comment.

remoteconfig/client.go
lahirumaramba: Reconsidered and this looks like a good first step. It is fine to leave this as it is. No need to make further changes. Please ignore my comment.

remoteconfig/client.go
lahirumaramba: Reconsidered and this looks like a good first step. It is fine to leave this as it is. No need to make further changes. Please ignore my comment.

remoteconfig/client.go
lahirumaramba: Reconsidered and this looks like a good first step. It is fine to leave this as it is. No need to make further changes. Please ignore the above comment.

remoteconfig/types.go
lahirumaramba: Same here. I think this is a good starting point. You can ignore my initial comment.

remoteconfig/client.go
allentv: I removed this

allentv:
@lahirumaramba Thanks for the feedback on how to fix those errors. I have made the changes and checks have passed. Appreciate your patience with this.

lahirumaramba (approved):
@allentv Thank you for fixing the build issues!
LGTM!

PR #403, opened 2020-09-30
Migrate SignBlob from IAM to IAMCredentials

kentengjin:
Point all SignBlob calls to iamcredentials.googleapis.com instead of iam.googleapis.com.


kentengjin:
Wrong branch. Closing pull.

PR #404, opened 2020-09-30
fix(auth): Migrate IAM SignBlob to IAMCredentials SignBlob

kentengjin:
Point all SignBlob to iamcredentials instead of iam.


hiranya911 (approved):
LGTM 👍

PR #408, opened 2020-11-02
fix: Ambiguous error message assertion

johejo:
In order not to break the test in multiple Go versions.

for https://github.com/golang/go/commit/6af088bfc66c13143c9ef46b4cf0805df77a8fbe


hiranya911:
Thanks for the PR @johejo. Did you actually encounter a test failure in some Go version? What prompted you to come up with this fix?

johejo:
This change is for 1.16. 
If we use Go tip, the test will fail. 
There is still time until 1.16 so we don't have to hurry.

johejo:
https://github.com/golang/go/commit/6af088bfc66c13143c9ef46b4cf0805df77a8fbe has reverted in https://github.com/golang/go/commit/283d65413db75edbc4691c4fecf23228509436f0

PR #413, opened 2020-12-17
Adding delayed response message for holidays

lahirumaramba:


lahirumaramba:
Merging this PR will cause `master` to go out of sync with `dev`, which could be problematic for future releases.

Issue #409, opened 2020-11-03
FR: Support Authentication Emulator


maku693:
### Step 2: Describe your environment

* Operating System version: all
* Firebase SDK version: 4.1.0
* Firebase Product: auth

### Step 3: Describe the problem

Firebase Local Emulator Suite has Authentication Emulator, but this SDK doesn't support the emulator yet.

While Realtime Database emulator support being discussed on #298, I believe emulator support can be implemented separately for each product, so I opened a new issue to discuss about Authentication Emulator support.

To support auth emulator, following implementation should be added (We can borrow the ideas of Node.js Admin SDK which already supports the emulator; cf. https://github.com/firebase/firebase-admin-node/pull/1044):

1. Provide a way to configure the auth client to use the Emulator Suite.
    - To be consistent with original Node.js implementation, auth emulator support should be enabled via setting `FIREBASE_AUTH_EMULATOR_HOST` environment variable.
2. Emulated signer for token generation / validation.
    - According to Node.js Admin SDK implementation, emulated signer returns null string as bytes from `Sign()`, and returns `firebase-auth-emulator@example.com` from `Email()`.
    - And also, we should add `Algorithm()` method to signers, to switch algorithm to sign custom tokens based on which signer the client uses (Emulated signer uses `none` algorithm, other signers use `RS256`. cf. https://github.com/firebase/firebase-admin-node/blob/dbb0c785c9b72590c19655d3ab5a39338e0cd393/src/auth/token-generator.ts#L231-L252).
3. Replace identity toolkit API endpoints with local (emulated) ones when using the emulator.

The steps above mostly been deduced from Node.js implementation, so they are incomplete perhaps.

#### Relevant Code:

I've submitted a PR. #414

samtstern:
@maku693 thank you for the detailed issue and the offer of a PR!  I think to start we'd definitely be interested in supporting `FIREBASE_AUTH_EMULATOR_HOST` and the JWT signing as well, but we should skip anything to do with JWT verification.  We're looking to improve the methods used in Node before porting those.

samtstern:
cc @yuchenshi

chris-rock:
Having this feature in this sdk would be really helpful!

kanekv:
This issue has "help-wanted" label, is it already being worked on or needs help/contribution?

yuchenshi:
This is on our roadmap, but we're still in the planning stage right now. We welcome contributions that help to land this feature sooner.

[...]

PR #414, opened 2020-12-18
feat(auth): Recognize `FIREBASE_AUTH_EMULATOR_HOST` environment variable

maku693:
If `FIREBASE_AUTH_EMULATOR_HOST` is set, auth API request will be sent to the host according to the environment variable.

### Discussion

#409 

~~Reflecting the [comment on the issue](https://github.com/firebase/firebase-admin-go/issues/409#issuecomment-743111881), the behavior of signers and token verifiers are unchanged.~~

Now the signer is replaced to `emulatedSigner`, and token verifiers panic when the environment variable is set. https://github.com/firebase/firebase-admin-go/pull/414#issuecomment-754931543 https://github.com/firebase/firebase-admin-go/pull/414#issuecomment-758166352

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -61,20 +61,15 @@ type Client struct {
 // Auth service through firebase.App.
 func NewClient(ctx context.Context, conf *internal.AuthConfig) (*Client, error) {
 	var (
-		signer cryptoSigner
-		err    error
+		isEmulator bool
+		signer     cryptoSigner
+		err        error
 	)
 
-	baseURL := defaultAuthURL
-	if authEmulatorHost := os.Getenv(emulatorHostEnvVar); authEmulatorHost != "" {
-		baseURL = fmt.Sprintf("http://%s/identitytoolkit.googleapis.com", authEmulatorHost)
+	authEmulatorHost := os.Getenv(emulatorHostEnvVar)
+	if authEmulatorHost != "" {
+		isEmulator = true
 	}
-
-	idToolkitV1Endpoint := fmt.Sprintf("%s/v1", baseURL)
-	idToolkitV2Beta1Endpoint := fmt.Sprintf("%s/v2beta1", baseURL)
-	userManagementEndpoint := idToolkitV1Endpoint
-	providerConfigEndpoint := idToolkitV2Beta1Endpoint
-	tenantMgtEndpoint := idToolkitV2Beta1Endpoint
 
 	creds, _ := transport.Creds(ctx, conf.Opts...)
 
@@ -105,12 +100,12 @@ func NewClient(ctx context.Context, conf *internal.AuthConfig) (*Client, error) 
 		}
 	}
 
-	idTokenVerifier, err := newIDTokenVerifier(ctx, conf.ProjectID)
+	idTokenVerifier, err := newIDTokenVerifier(ctx, conf.ProjectID, isEmulator)
 	if err != nil {
 		return nil, err
 	}
 
-	cookieVerifier, err := newSessionCookieVerifier(ctx, conf.ProjectID)
+	cookieVerifier, err := newSessionCookieVerifier(ctx, conf.ProjectID, isEmulator)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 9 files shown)


auth/auth_test.go
```diff
@@ -280,6 +283,35 @@ func TestNewClientExplicitNoAuth(t *testing.T) {
...
+func TestNewClientEmulatorHostEnvVar(t *testing.T) {
+	os.Setenv(emulatorHostEnvVar, "localhost:9099")
+	defer os.Unsetenv(emulatorHostEnvVar)
+
+	conf := &internal.AuthConfig{
+		Opts: []option.ClientOption{
+			option.WithoutAuthentication(),
+		},
+	}
+	client, err := NewClient(context.Background(), conf)
```
maku693, line 299: I'm not sure using no-auth config is appropriate for this test.

auth/auth_test.go
```diff
@@ -1163,23 +1195,26 @@ func checkCookieVerifier(tv *tokenVerifier, projectID string) error {
...
+	baseClient := client.baseClient
+	if baseClient.userManagementEndpoint != defaultIDToolkitV1Endpoint {
+		return fmt.Errorf("userManagementEndpoint = %q; want = %q", baseClient.userManagementEndpoint, defaultIDToolkitV1Endpoint)
+	}
+	if baseClient.providerConfigEndpoint != defaultIDToolkitV2Beta1Endpoint {
+		return fmt.Errorf("providerConfigEndpoint = %q; want = %q", baseClient.providerConfigEndpoint, defaultIDToolkitV2Beta1Endpoint)
 	}
-	if umc.providerConfigEndpoint != providerConfigEndpoint {
-		return fmt.Errorf("providerConfigEndpoint = %q; want = %q", umc.providerConfigEndpoint, providerConfigEndpoint)
+	if baseClient.tenantMgtEndpoint != defaultIDToolkitV2Beta1Endpoint {
```
maku693, line 1205: Added the check for `tenantMgtEndpoint`.

samtstern:
General approach LGTM but assigning to @hiranya911 for Go / Admin SDK review.

auth/auth.go
```diff
@@ -62,6 +64,17 @@ func NewClient(ctx context.Context, conf *internal.AuthConfig) (*Client, error)
 		err    error
 	)
 
+	baseURL := defaultAuthURL
+	if authEmulatorHost := os.Getenv("FIREBASE_AUTH_EMULATOR_HOST"); authEmulatorHost != "" {
```
hiranya911, line 68: Should probably define a local constant for the env variable name.

hiranya911 (approved):
LGTM. Only question is should we release this without support for emulating token verification?

dgpc:
@hiranya911 - as a user waiting for this feature, my 2c worth is that although token verification emulation is necessary for complete end-to-end testing, making it possible to talk to the emulator at all in this manner is still useful to release as-is, since it unblocks testing of a variety of workflows

Michaelhobo:
> @hiranya911 - as a user waiting for this feature, my 2c worth is that although token verification emulation is necessary for complete end-to-end testing, making it possible to talk to the emulator at all in this manner is still useful to release as-is, since it unblocks testing of a variety of workflows

Agreed, this would be a huge win for me.

yuchenshi:
I'd propose a middle ground where we release a version first, where everything works but token verification panics with "not implemented" for now when the env var is set. This should ensure that SDK don't call production APIs in emulator mode and also minimize any confusion caused by the SDK verifying signatures using production public keys.

samtstern:
@yuchenshi I like that idea!  SGTM

hiranya911:
Yeah, I like that too. Any chance we can get this PR updated to incorporate that change?

maku693:
Thank you for review! I'll update this PR this weekend.

auth/token_verifier.go
```diff
@@ -100,9 +100,10 @@ type tokenVerifier struct {
 	expiredTokenCode  string
 	keySource         keySource
 	clock             internal.Clock
+	isEmulator        bool
```
maku693, line 103: This might be too simple. It's better to implement custom `keySource` for emulator mode. 

But I don't sure custom `keySource` is necessary because
- We didn't decide how to verify tokens in the emulator mode yet.
- Cannot skip verification runs before signature verification.

auth/auth.go
maku693: I'd define a constant. https://github.com/firebase/firebase-admin-go/pull/414/commits/11d9970757fb1ee5e6ff0992ed219d8bfd4cbdc0#diff-f5c9a39fac1719f2a73593994b931d8a874449548e8a3753e1be46e761e503eaR33

maku693:
@yuchenshi Is the point of the change is to prevent calling production APIs in emulator mode?
If so, not just the token verifier but also the signer and the tenant manager should panic at API calls, correct?

yuchenshi:
For `createCustomToken` etc., when `FIREBASE_AUTH_EMULATOR_HOST` is set, the SDK should create unsigned tokens (i.e. the signature part being an empty string), which are in turn accepted by the Auth Emulator. In other words, signer should be short-circuited and there should not be any outgoing network call at all (no `signBlob` etc.).

Auth Emulator doesn't support multi-tenancy (at least for now) and I'm indifferent whether the SDK should panic or forward them to the Auth Emulator which rejects such requests. I'm fine as long as they won't reach production.

maku693:
Ah, I understand. I'll update not to sign tokens.

hiranya911 (approved):
Looks reasonable to me. 

@yuchenshi wdyt?

yuchenshi (approved):

PR #418, opened 2021-02-03
Lint

chemidy:
fix omitempty


chemidy:
@googlebot I fixed it.

PR #419, opened 2021-02-20
feat(auth): Add token verification logic for emulator mode

maku693:
### Discussion

#409

This PR adds token verification with auth emulator for auth client, in addition to the changes introduced at #414.


maku693:
We should add integration tests for this PR, like Node.js SDK implementation. https://github.com/firebase/firebase-admin-node/pull/1148

But I'm not sure how much should we add tests for this feature, and how to setup CI. any suggestions?

auth/auth.go
```diff
@@ -338,6 +342,11 @@ func (c *baseClient) VerifyIDTokenAndCheckRevoked(ctx context.Context, idToken s
 	return decoded, nil
 }
 
+// IsTenantIDMismatch checks if the given error was due to a mismatched tenant ID in a JWT.
```
maku693, line 345: This is not changed, just moved to improve readability.

auth/auth_test.go
```diff
@@ -697,25 +698,88 @@ func TestVerifyIDTokenWithNoProjectID(t *testing.T) {
 	}
 }
 
-func TestVerifyIDTokenInEmulatorMode(t *testing.T) {
-	os.Setenv(emulatorHostEnvVar, "localhost:9099")
-	defer os.Unsetenv(emulatorHostEnvVar)
+func TestVerifyIDTokenUnsigned(t *testing.T) {
```
maku693, line 701: Also added a test to guarantee that the non-emulator mode implementation doesn't allow unsigned tokens.

auth/token_verifier.go
```diff
@@ -269,13 +267,13 @@ func (tv *tokenVerifier) verifyHeaderAndBody(token string) (*Token, error) {
...
 	issuer := tv.issuerPrefix + tv.projectID
-	if header.KeyID == "" {
+	if !isEmulator && header.KeyID == "" {
 		if payload.Audience == firebaseAudience {
 			return nil, fmt.Errorf("expected %s but got a custom token", tv.articledShortName)
 		}
 		return nil, fmt.Errorf("%s has no 'kid' header", tv.shortName)
 	}
-	if header.Algorithm != "RS256" {
+	if !isEmulator && header.Algorithm != "RS256" {
```
maku693, line 276: This is different from [Node.js implementation](https://github.com/firebase/firebase-admin-node/blob/93362d5721fc45f3fc541543813d5739eda8ba2d/src/auth/token-verifier.ts#L200) but I believe this is OK because we can skip algorithm verification in emulator mode and the feature that can specify an algorithm for token verification is not required for this Go SDK.

auth/auth.go
```diff
@@ -281,43 +283,45 @@ func (c *baseClient) withTenantID(tenantID string) *baseClient {
...
 func (c *baseClient) VerifyIDTokenAndCheckRevoked(ctx context.Context, idToken string) (*Token, error) {
-	decoded, err := c.VerifyIDToken(ctx, idToken)
+	return c.verifyIDToken(ctx, idToken, true)
+}
+
+func (c *baseClient) verifyIDToken(ctx context.Context, idToken string, checkRevoked bool) (*Token, error) {
+	decoded, err := c.idTokenVerifier.VerifyToken(ctx, idToken, c.isEmulator)
 	if err != nil {
-		return nil, err
+		return decoded, err
```
hiranya911, line 310: return nil, err

auth/auth.go
```diff
@@ -281,43 +283,45 @@ func (c *baseClient) withTenantID(tenantID string) *baseClient {
...
+		return nil, &internal.FirebaseError{
+			ErrorCode: internal.InvalidArgument,
+			String:    fmt.Sprintf("invalid tenant id: %q", decoded.Firebase.Tenant),
+			Ext: map[string]interface{}{
+				authErrorCode: tenantIDMismatch,
+			},
+		}
+	}
+
+	if !c.isEmulator && !checkRevoked {
```
hiranya911, line 323: Nit: Might be slightly more readable if we flipped this condition around:

```go
if c.isEmulator || checkRevoked {
  revoked, err := c.checkRevoked(ctx, decoded)
  // ...
}

return decoded, nil
```

auth/auth.go
```diff
@@ -352,27 +361,35 @@ func IsIDTokenRevoked(err error) bool {
...
+// In non-emulator mode, this function does not make any RPC calls most of the time.
+// The only time it makes an RPC call is when Google public keys need to be refreshed.
+// These keys get cached up to 24 hours, and therefore the RPC overhead gets amortized
+// over many invocations of this function.
 //
 // This does not check whether or not the cookie has been revoked. Use `VerifySessionCookieAndCheckRevoked()`
 // when a revocation check is needed.
 func (c *Client) VerifySessionCookie(ctx context.Context, sessionCookie string) (*Token, error) {
-	return c.cookieVerifier.VerifyToken(ctx, sessionCookie)
+	return c.verifySessionCookie(ctx, sessionCookie, false)
```
hiranya911, line 372: @yuchenshi do we support emulator mode for session cookies?

auth/auth.go
```diff
@@ -352,27 +361,35 @@ func IsIDTokenRevoked(err error) bool {
...
+}
+
+func (c *Client) verifySessionCookie(ctx context.Context, sessionCookie string, checkRevoked bool) (*Token, error) {
+	decoded, err := c.cookieVerifier.VerifyToken(ctx, sessionCookie, c.isEmulator)
 	if err != nil {
-		return nil, err
+		return decoded, err
+	}
+
+	if !c.isEmulator && !checkRevoked {
```
hiranya911, line 391: Consider flipping the condition as above.

auth/token_verifier.go
hiranya911: I think in emulator mode we should look for the algorithm `none`. @yuchenshi your thoughts?

auth/auth_test.go
```diff
@@ -697,25 +698,88 @@ func TestVerifyIDTokenWithNoProjectID(t *testing.T) {
...
 
-	defer func() {
-		if r := recover(); r == nil {
-			t.Error("VeridyIDToken() didn't panic")
-		}
-	}()
-	c.VerifyIDToken(context.Background(), testIDToken)
+	token := getEmulatedIDToken(nil)
+	_, err = client.VerifyIDToken(context.Background(), token)
+	if err == nil || !strings.HasPrefix(err.Error(), "failed to establish a connection") {
```
hiranya911, line 780: Can we also test for the error code in this scenario? What's the error code that gets set in this situation?

hiranya911 (changes requested):
Thanks for the PR @maku693. This looks pretty good overall. I've left a few comments on improving the readability a bit. Plus a couple of product questions to @yuchenshi around session cookie verification support.

auth/auth.go
yuchenshi: Yes, the emulator has full support for (unsigned, fake) session cookies

auth/token_verifier.go
yuchenshi: According to the design doc approved, the header should not be checked at all when `isEmulator` (to allow future changes). Therefore @maku693's implementation is correct and afaict is the same as the Node.js implementation. @maku693 would you mind explaining why you think it would be different from the Node.js implementation?

maku693: @yuchenshi Ah, sorry about my confusing comment. I mean ”This is not exactly same as what Node.js SDK does”. Node.js implementation does checks algorithm, regardless of if it is in emulator mode or not, like @hiranya911 says.
So in emulator mode, this is basically same as Node.js implementation.

maku693:
Thank you for reviewing! I've fixed the points you've commented.

auth/token_verifier.go
hiranya911: Looks like  Node.js skips algorithm check in emulator mode now. So this is fine.

https://github.com/firebase/firebase-admin-node/blob/93362d5721fc45f3fc541543813d5739eda8ba2d/src/auth/token-verifier.ts#L200-L203

hiranya911 (approved):
I'm at an LGTM.

@yuchenshi do you have any more feedback?

yuchenshi (approved):

robinmitra:
Would be awesome to get this merged in if it all looks good, as this is breaking my local development flow.

maku693:
Hi @hiranya911 and @yuchenshi, do you have any plans to release this in the near future?

hiranya911:
I've merged the PR. It will be included in the next release (possibly in another week or so).

zkrzyzanowski:
Hi all,
Thank you to those that worked on this and got this merged in! Is there an ETA on when the release with this in it might happen? It'd be great to be able to use the `admin` sdk with the local emulator, as this is breaking some of my dev workflow at the moment. Any other suggestions would be appreciated if there's a workaround. 

Thank you!

hiranya911:
Will try and get it released this week.

saulortega:
Hi.

@maku693 @hiranya911 

Why when using emulator it ALWAYS checks if token was revoked? It makes no sense to me.

image

maku693:
@saulortega 

I just implemented similar logic based on the Node.js implementation as a reference, so I don't have an exact understanding of the reasons.

However, according to the [emulator's documentation](https://firebase.google.com/docs/emulator-suite/connect_auth?hl=en#custom_token_authentication), it seems that the emulator itself does not verify the expiration of custom tokens. Therefore, aligning with that behavior could be a possible reason.

Issue #421, opened 2021-03-02
FR: MFA support


jasperkuperus:
### Describe your environment

  * Operating System version: All
  * Firebase SDK version: v4.2.0
  * Firebase Product: auth

### Describe the problem

It would be nice if this library had support for MFA. Specifically, I'd like to get the enrolled second factors for a given user. For Node.js this is [available and documented](https://cloud.google.com/identity-platform/docs/admin/manage-mfa-users), but I don't see such a feature in this Go library. Are there any plans to add this?

hiranya911:
No specific plans yet. But we are accepting PRs if somebody wishes to help speed things along.

jasperkuperus:
Unfortunately, I don't have time to fully implement MFA. But I did just open a PR to add the MFA info to the user record. This already adds great value in my situation.

hiranya911:
PR has been merged and updated to meet our API design requirements. This will be included in the next release.

PR #422, opened 2021-03-09
feat(auth): Add MFA info to UserRecord

jasperkuperus:
### Discussion

Issue: https://github.com/firebase/firebase-admin-go/issues/421

This PR implements a small part of Multi Factor Authentication (MFA). I do realize this isn't a full implementation, unfortunately I don't have time for that. This PR only implements unmarshalling the MFA info from the REST response into the `UserRecord` struct.

For structure and naming within the `UserRecord`, I've decided to be consistent with the Node.js client library, which already has this implemented: https://firebase.google.com/docs/reference/admin/node/admin.auth.UserRecord

### Testing

  * I've updated the existing tests to also incorporate MFA. On the existing `testUser`, I've added MFA info. Next to that, I've introduced a test user without MFA. In the list endpoints, I replaced every third user with a non MFA user, so the list tests both users with and without MFA info available. All unit tests should be OK.
  * Unfortunately, I wasn't able to get the integration tests running locally, so I'm not sure whether I did break something there. I did manage to test the code in an actual project with a small local setup and verified that it works.

### API Changes

  * This PR does not change the public API. The REST API already returns the `mfaInfo` section. This PR merely adds the code to pick it up into Golang structs.


auth/user_mgt.go
```diff
@@ -56,6 +56,19 @@ type UserInfo struct {
 	UID        string `json:"rawId,omitempty"`
 }
 
+// MultiFactorInfo describes a user enrolled second factor.
+type MultiFactorInfo struct {
```
hiranya911, line 60: This API should closely mirror the type signatures defined for Node.js:

```go
type MultiFactorInfo struct {
  UID string
  DisplayName string
  EnrollmentTimestamp int64
  FactorID string
}
```

You will probably need to define a multiFactorInfoQuery internal type for unmarshalling, and then initialize this type based on that.

hiranya911 (changes requested):
Thanks @jasperkuperus for the PR. It looks mostly good. I've suggested some changes to the public API.

@bojeil-google how do you feel about just releasing the changes to `UserRecord` without the support for updating MFA settings?

bojeil-google:
> Thanks @jasperkuperus for the PR. It looks mostly good. I've suggested some changes to the public API.
> 
> @bojeil-google how do you feel about just releasing the changes to `UserRecord` without the support for updating MFA settings?

I think it's fine to do so, as long as we keep track of the missing functionality (updating MFA settings). Ultimately, we want to make sure we have feature parity across all languages.

auth/user_mgt.go
jasperkuperus: Ah, I must've halfway through accidentally switched to mapping names to the REST API in stead of the Node.js API. I see. I've had a look at how the Node.js library converts the REST response JSON to the internal structures. I'll fix this early next week!

jasperkuperus: Mental note: https://github.com/firebase/firebase-admin-node/blob/f624b2963900deebed0fdb5cbe66ce7eb0d244b8/src/auth/user-record.ts

jasperkuperus: Instead of what you proposed `multiFactorInfoQuery`, I've sticked with `MultiFactorInfoResponse`, which makes more sense to me and is in line with the Node.js naming.

auth/user_mgt.go
```diff
@@ -56,6 +56,37 @@ type UserInfo struct {
 	UID        string `json:"rawId,omitempty"`
 }
 
+// MultiFactorInfoResponse describes the `mfaInfo` of the user record API response
+type MultiFactorInfoResponse struct {
+	MFAEnrollmentID string `json:"mfaEnrollmentId"`
```
hiranya911, line 61: Suggest also adding `omitempty` to everything (just for consistency with the rest of this file).

testdata/list_users.json
```diff
@@ -63,21 +71,29 @@
             "createdAt": "1234567890000",
             "lastLoginAt": "1233211232000",
             "customAttributes": "{\"admin\": true, \"package\": \"gold\"}",
-            "tenantId": "testTenant"
+            "tenantId": "testTenant",
+            "mfaInfo": [
+              {
+                "phoneInfo": "+31612345678",
```
hiranya911, line 77: Nit: Suggest using `+1234567890` as the phone number and `testuser` as the enrollment ID to make it obvious that they are test data.

testdata/get_user.json
```diff
@@ -31,7 +31,15 @@
       "createdAt": "1234567890000",
       "lastLoginAt": "1233211232000",
       "customAttributes": "{\"admin\": true, \"package\": \"gold\"}",
-      "tenantId": "testTenant"
+      "tenantId": "testTenant",
+      "mfaInfo": [
+        {
+          "phoneInfo": "+31612345678",
```
hiranya911, line 37: Here too. Use obviously fake data.

hiranya911 (changes requested):
Thanks for making the changes @jasperkuperus. This is starting to shape up nicely. Just a few more changes needed to get the public API sorted out.

auth/user_mgt.go
```diff
@@ -56,6 +56,37 @@ type UserInfo struct {
...
+	UID                 string
+	DisplayName         string
+	EnrollmentTimestamp int64
+	FactorID            MultiFactorID
+	PhoneNumber         string
+}
+
+// MultiFactorSettings describes the multi-factor related user settings.
+type MultiFactorSettings struct {
+	EnrolledFactors []*PhoneMultiFactorInfo
```
hiranya911, line 87: This cannot be `PhoneMultiFactorInfo`. We'd want to support other auth types in the future, and at that point we'll have to break this API. Let's just rename `PhoneMultiFactorInfo` struct to `MultiFactorInfo`. Then we can add new fields to that type when we introduce a new auth factor, and that won't be a breaking change.

hiranya911 (changes requested):

hiranya911 (approved):
Thanks @jasperkuperus. This LGTM 👍 

I do need to get this API approved through our internal API review process. I'll merge when I have the required approvals.

hiranya911:
API review process for this has been initiated. Should have some updates by next week.

hiranya911:
This API has been approved with some minor changes. The change being removal of the `MultiFactorID` type and the related `Phone` constant (we will use `FactorID string` instead).

@jasperkuperus would you like to push the change to this PR? If you're busy I can merge this as-is, and implement the required changes on top of that.

jasperkuperus:
> @jasperkuperus would you like to push the change to this PR? If you're busy I can merge this as-is, and implement the required changes on top of that.

@hiranya911  It's a small change, but I have a rather busy week for me. If you could be so kind to merge it and implement the change on top of that, it'd be of great help, thanks! 😊

Issue #415, opened 2021-01-13
Unable to connect to realtime db hosted in europe west


Jo3116:
* Operating System version: macOS Catalina 10.15.6
  * Firebase SDK version: 4.1.0
  * Library version: Unsure Latest i believe
  * Firebase Product: realtime database

###The problem
Cannot connect to a realtime Db hosted in Europe west because of invalid database URL. I was attempting to use a url ending in `europe-west1.firebasedatabase.app` to connect to a Europe hosting but received the error `invalid database URL: want host: "firebaseio.com` 

### Steps to reproduce:
Attempt to create a new client using `app.DatabaseWithURL` passing in a Europe url ending in `europe-west1.firebasedatabase.app`

### Relevant Code:

```// NewFirebaseConnection creates a new instance of the FirebaseConnection Struct
func NewFirebaseConnection(ctx context.Context, url string) FirebaseConnection {
	var conn FirebaseConnection
	conn.app = initializeApp(ctx)
	conn.client = initializeClient(ctx, conn.app, url)
	return conn
}

// initializeApp creates a firebase app object
func initializeApp(ctx context.Context) *firebase.App {
	// Initialize the app with a service account, granting admin privileges
	app, err := firebase.NewApp(ctx, nil)
	if err != nil {
		log.Fatalln("Error initializing app:", err)
	}

	return app
}

// initializeClient creates a firebase client object
func initializeClient(ctx context.Context, app *firebase.App, url string) *db.Client {
	client, err := app.DatabaseWithURL(ctx, url)
	if err != nil {
		log.Fatalln("Error initializing database client:", err)
	}

	return client
}

hiranya911:
This requires a fix similar to https://github.com/firebase/firebase-admin-python/pull/517

thameezb:
Hitting the same issue

PR #423, opened 2021-03-13
fix(rtdb): enable support for firebasedatabase.app

thameezb:
Closes https://github.com/firebase/firebase-admin-go/issues/415

Enables support for Firebase DBs hosted in the EU


thameezb:
> @googlebot I signed it!

db/db.go
```diff
@@ -48,8 +48,8 @@ func NewClient(ctx context.Context, c *internal.DatabaseConfig) (*Client, error)
 		return nil, err
 	} else if p.Scheme != "https" {
 		return nil, fmt.Errorf("invalid database URL: %q; want scheme: %q", c.URL, "https")
-	} else if !strings.HasSuffix(p.Host, ".firebaseio.com") {
-		return nil, fmt.Errorf("invalid database URL: %q; want host: %q", c.URL, "firebaseio.com")
+	} else if !(strings.HasSuffix(p.Host, ".firebaseio.com") || strings.HasSuffix(p.Host, ".firebasedatabase.app")) {
```
hiranya911, line 51: I think going forward we want to remove this domain validation entirely. That's what we did in our Python SDK. Can we change this PR to just drop the domain validation?

hiranya911 (changes requested):
Sorry for the delayed response. This somehow fell through the cracks. I just had one suggestion. And please rebase agains the dev branch. We don't directly push to the default branch.

db/db.go
thameezb: Done

thameezb (commented):
Rebased onto dev

hiranya911 (approved):
Thanks. LGTM 👍

PR #436, opened 2021-05-03
fix(auth): adds missing EMAIL_NOT_FOUND error code

bojeil-google:
Catch EMAIL_NOT_FOUND when /accounts:sendOobCode is called for password reset on a user that does not exist.


auth/user_mgt.go
```diff
@@ -468,6 +469,11 @@ func IsEmailAlreadyExists(err error) bool {
 	return hasAuthErrorCode(err, emailAlreadyExists)
 }
 
+// IsEmailNotFound checks if the given error was due to the user record corresponding to the email not being found.
```
hiranya911, line 472: Lets also add a test case at https://github.com/firebase/firebase-admin-go/blob/4121c50b610f0ce408a2978cf14b306becf276d3/auth/user_mgt_test.go#L510

hiranya911 (changes requested):
Looks good. Please change the target branch to `dev`. And lets also add a test case.

auth/user_mgt.go
bojeil-google: I added a test to `email_action_links_test.go` as that is where all the `PasswordResetLinkWithSettings` tests are located.

bojeil-google:
> Looks good. Please change the target branch to `dev`. And lets also add a test case.

Done

hiranya911 (approved):

auth/user_mgt.go
egilmorez, line 472: Is this comment parsed into reference docs?  If so, we would probably want to make the literal `IsEmailNotFound` code font (not sure if backticks work for Go).

We could do that separately, since the same would apply for IsInsufficientPermission and any other similar things.

auth/email_action_links_test.go
```diff
@@ -177,6 +179,27 @@ func TestPasswordResetLinkWithSettings(t *testing.T) {
...
+	s := echoServer([]byte(resp), t)
+	defer s.Close()
+	s.Status = http.StatusBadRequest
+
+	link, err := s.Client.PasswordResetLinkWithSettings(context.Background(), testEmail, testActionCodeSettings)
+	if link != "" || err == nil {
+		t.Errorf("PasswordResetLinkWithSettings() = (%q, %v); want = (%q, error)", link, err, "")
+	}
+
+	want := "no user record found for the given email"
```
egilmorez, line 197: Is this the error string shown in an error message?

In the other langs we made these messages "grammatical" with caps and punctuation, like "No user record found for the given email."

auth/user_mgt.go
```diff
@@ -1298,6 +1304,11 @@ var serverError = map[string]*authError{
 		message:  "user with the provided email already exists",
 		authCode: emailAlreadyExists,
 	},
+	"EMAIL_NOT_FOUND": {
+		code:     internal.NotFound,
+		message:  "no user record found for the given email",
```
egilmorez, line 1309: Is this the error string shown in an error message?

In the other langs we made these messages "grammatical" with caps and punctuation, like "No user record found for the given email."

egilmorez (approved):
LG, some nits we can fix here or in a doc fixit later.

auth/user_mgt.go
bojeil-google: These are visible externally. All other functions follow the same pattern (without backticks):
https://pkg.go.dev/firebase.google.com/go@v3.13.0+incompatible/auth#IsConfigurationNotFound
I am not sure if backticks work for Go either.
Will need to confirm with @hiranya911

hiranya911: This is consistent with godoc conventions. Comment should start with the name of the function/type name without any formatting. See https://blog.golang.org/godoc and https://golang.org/src/go/doc/example.go

auth/email_action_links_test.go
bojeil-google: This is the same error as below. Gonna mark it as duplicate.

auth/user_mgt.go
bojeil-google: I agree with you in general but it seems like all error messages follow this pattern in Go: 
https://github.com/firebase/firebase-admin-go/blob/4121c50b610f0ce408a2978cf14b306becf276d3/auth/user_mgt.go#L1280
Will need to confirm with @hiranya911

auth/user_mgt.go
bojeil-google: Thanks @hiranya911

auth/email_action_links_test.go
hiranya911: This is again consistent with Go conventions: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

bojeil-google: Thanks for confirming.

auth/user_mgt.go
bojeil-google: Marking this as resolved as this is consistent with Go conventions: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

PR #439, opened 2021-05-17
chore: Installing golint before checkout

hiranya911:
* Restricting build matrix to the oldest and latest go versions we support.
* Installing `golint` before checking out source to avoid accidental changes to the module dependencies (`go.mod` and `go.sum` files).
* Bumped the go version to 1.16 for the GOPATH build, since in this mode go pulls all the dependencies at their HEAD versions, and some of them no longer works on go 1.12.


lahirumaramba (approved):
Thanks!

Issue #443, opened 2021-05-26
Incorrect import of user metadata


valery-barysok:
Library version: Latest

During the import of users I need to provide partial metadata for user to import but it requires both.

My code looks like:

```go
var metadata auth.UserMetadata
if user.CreatedAt.Valid {
	metadata.CreationTimestamp = user.CreatedAt.Time.Unix() * 1000
}

if user.LastLogInAt.Valid {
	metadata.LastLogInTimestamp = user.LastLogInAt.Time.Unix() * 1000
}

users = append(users,
	(&auth.UserToImport{}).
		UID(...).
		Email(...).
		EmailVerified(...).
		Metadata(&metadata).
		PasswordHash(...),
)

...

result, err := authClient.ImportUsers(ctx, users, opts))

...

```

PR #444, opened 2021-05-26
Fix incorrect import of user metadata.

valery-barysok:
It fixes the import issue described in https://github.com/firebase/firebase-admin-go/issues/443


valery-barysok:
@googlebot I signed it!

hiranya911:
PR should be sent against the dev branch. And also please add some tests to verify this behavior.

PR #455, opened 2021-08-12
fix(auth): check disabled status on verifyIDToken

Galadros:
The `verifyIDToken()` function checks to see whether the ID token is revoked, but never actually checks to see whether or not the user is disabled.  This change makes it so that `verifyIDToken()` does check the `disabled` field by modifying `checkRevoked()`.  The function is renamed to `checkRevokedOrDisabled` and now additionally includes a check to make sure that the user is not disabled.  A new `authErrorCode` corresponding to this case is added.  `verifySessionCookie()` is slightly modified to account for the fact that `checkRevokedOrDisabled` now returns 3 values instead of 2.  `verifyIDToken()` has been modified to raise an `InvalidArgument` error with `authErrorCode` of "ID_TOKEN_DISABLED" if the user is disabled.


auth/auth.go
```diff
@@ -37,6 +37,7 @@ const (
 
 	// SDK-generated error codes
 	idTokenRevoked       = "ID_TOKEN_REVOKED"
+	idTokenDisabled      = "ID_TOKEN_DISABLED"
```
hiranya911, line 40: This is called `USER_DISABLED` in other languages. Let's use the same terminology here.

auth/auth.go
```diff
@@ -335,6 +335,15 @@ func (c *baseClient) verifyIDToken(ctx context.Context, idToken string, checkRev
 				},
 			}
 		}
+		if disabled {
```
hiranya911, line 338: Let's check for this before revoked.

auth/auth.go
```diff
@@ -335,6 +335,15 @@ func (c *baseClient) verifyIDToken(ctx context.Context, idToken string, checkRev
...
+			return nil, &internal.FirebaseError{
+				ErrorCode: internal.InvalidArgument,
+				String:    "ID token has been disabled",
```
hiranya911, line 341: User account has been disabled.

auth/auth.go
```diff
@@ -347,11 +356,18 @@ func IsTenantIDMismatch(err error) bool {
 
 // IsIDTokenRevoked checks if the given error was due to a revoked ID token.
 //
-// When IsIDTokenRevoked returns true, IsIDTokenInvalid is guranteed to return true.
+// When IsIDTokenRevoked returns true, IsIDTokenInvalid is garuanteed to return true.
```
hiranya911, line 359: guaranteed

auth/auth.go
```diff
@@ -347,11 +356,18 @@ func IsTenantIDMismatch(err error) bool {
...
 func IsIDTokenRevoked(err error) bool {
 	return hasAuthErrorCode(err, idTokenRevoked)
 }
 
+// IsIDTokenDisabled checks if the given error was due to a disabled ID token
+//
+// When IsIDTokenDisabled returns true, IsIDTokenInvalid is garuanteed to return true.
+func IsIDTokenDisabled(err error) bool {
```
hiranya911, line 367: IsUserDisabled

auth/auth.go
```diff
@@ -408,18 +425,24 @@ func (c *Client) verifySessionCookie(ctx context.Context, sessionCookie string,
...
 //
-// When IsSessionCookieRevoked returns true, IsSessionCookieInvalid is guranteed to return true.
+// When IsSessionCookieRevoked returns true, IsSessionCookieInvalid is guaranteed to return true.
 func IsSessionCookieRevoked(err error) bool {
 	return hasAuthErrorCode(err, sessionCookieRevoked)
 }
 
-func (c *baseClient) checkRevoked(ctx context.Context, token *Token) (bool, error) {
+// checkRevokedOrDisabled checks whether the input token has been revoked or disabled.
+func (c *baseClient) checkRevokedOrDisabled(ctx context.Context, token *Token) (revoked bool, disabled bool, err error) {
```
hiranya911, line 434: Instead of expanding the return signature, it's probably better to just return a single error value from this function.

```
checkRevokedOrDisabled(ctx context.Context, token *Token, revokedErrCode string) err error
```

Then the caller can be just:

```go
if c.isEmulator || checkRevokedOrDisabled {
		if err := c.checkRevokedOrDisabled(ctx, decoded, idTokenRevoked); err != nil {
			return nil, err
		}
}
```

auth/auth.go
```diff
@@ -408,18 +425,24 @@ func (c *Client) verifySessionCookie(ctx context.Context, sessionCookie string,
...
 	if err != nil {
-		return false, err
+		return false, false, err
 	}
-
-	return token.IssuedAt*1000 < user.TokensValidAfterMillis, nil
+	if token.IssuedAt*1000 < user.TokensValidAfterMillis {
+		return true, false, nil
+	}
+	if user.Disabled {
```
hiranya911, line 442: Check for disabled first.

auth/auth_test.go
```diff
@@ -858,7 +858,7 @@ func TestInvalidTokenDoesNotCheckRevoked(t *testing.T) {
 	s.Client.idTokenVerifier = testIDTokenVerifier
 
 	ft, err := s.Client.VerifyIDTokenAndCheckRevoked(context.Background(), "")
-	if ft != nil || !IsIDTokenInvalid(err) || IsIDTokenRevoked(err) {
+	if ft != nil || !IsIDTokenInvalid(err) || IsIDTokenRevoked(err) || IsIDTokenDisabled(err) {
```
hiranya911, line 861: Please add test cases to cover the new functionality.

snippets/auth.go
```diff
@@ -1426,7 +1426,11 @@ func verifyIDTokenAndCheckRevokedTenant(ctx context.Context, tenantClient *auth.
 		if auth.IsIDTokenRevoked(err) {
 			// Token is revoked. Inform the user to reauthenticate or signOut() the user.
 		} else {
-			// Token is invalid
+			if auth.IsIDTokenDisabled(err) {
```
hiranya911, line 1429:
```go
if auth.IsIDTokenRevoked(err) {
			// Token is revoked. Inform the user to reauthenticate or signOut() the user.
} else if auth.IsUserDisabled(err) {
                        // User is disabled
} else {
                       // Token is invalid
}		
```

hiranya911 (changes requested):
Pointed out a several things that need to change. Please change the base branch of the PR to `dev`. We don't directly merge to `master` on this repo.

Galadros:
I'm not sure how to fix the google-cla bot error.  It's complaining that "google-oss-bot" doesn't have a CLA signed.

hiranya911:
You might want to rebase your commits at dev and do a force push, which should clean up the commit history.

auth/auth.go
```diff
@@ -322,27 +322,28 @@ func (c *baseClient) verifyIDToken(ctx context.Context, idToken string, checkRev
 	}
 
 	if c.isEmulator || checkRevokedOrDisabled {
-		revoked, disabled, err := c.checkRevokedOrDisabled(ctx, decoded)
+		errMessageDisabled := &internal.FirebaseError{
+			ErrorCode: internal.InvalidArgument,
+			String:    "User has been disabled",
```
hiranya911, line 327: Go convention is to start off error messages with lowercase letters (unless acronym or other shorthand like "ID").

auth/auth.go
```diff
@@ -320,20 +321,29 @@ func (c *baseClient) verifyIDToken(ctx context.Context, idToken string, checkRev
...
-				},
-			}
+		errMessageRevoked := &internal.FirebaseError{
+			ErrorCode: internal.InvalidArgument,
+			String:    "ID token has been revoked",
+			Ext: map[string]interface{}{
+				authErrorCode: idTokenRevoked,
+			},
+		}
+		err = c.checkRevokedOrDisabled(ctx, decoded, idTokenRevoked, errMessageRevoked)
```
hiranya911, line 344: This seems to be calling `checkRevokedOrDisabled()` twice, which is not correct. That results in extra RPCs that we don't need. See my comment on how to refactor the `checkRevokedOrDisabled()` function, so this can be avoided.

auth/auth.go
```diff
@@ -380,46 +397,49 @@ func (c *Client) VerifySessionCookieAndCheckRevoked(ctx context.Context, session
...
 //
-// When IsSessionCookieRevoked returns true, IsSessionCookieInvalid is guranteed to return true.
+// When IsSessionCookieRevoked returns true, IsSessionCookieInvalid is guaranteed to return true.
 func IsSessionCookieRevoked(err error) bool {
 	return hasAuthErrorCode(err, sessionCookieRevoked)
 }
 
-func (c *baseClient) checkRevoked(ctx context.Context, token *Token) (bool, error) {
+// checkRevokedOrDisabled checks whether the input token has been revoked or disabled.
+func (c *baseClient) checkRevokedOrDisabled(ctx context.Context, token *Token, errCode string, errMessage *internal.FirebaseError) error {
```
hiranya911, line 431: This should look something like:

```go
func (c *Client) checkRevokedOrDisabled(ctx context.Context, token *Token, errCode, errMessage string) error {
        user, err := c.GetUser(ctx, token.UID)
	if err != nil {
		return err
	}

        if user.Disabled {
		// Create and return FirebaseError with code USER_DISABLED
	}

        if token.IssuedAt*1000 < user.TokensValidAfterMillis {
            // return FirebaseError constructed from errCode and errMessage
        }

        return nil
}
```

testdata/get_disabled_user.json
```diff
@@ -0,0 +1,45 @@
+{
```
hiranya911, line 1: Nit: Since this is just about testing a disabled user record, I'd say remove fields like `providerUserInfo` and `mfaInfo`.

auth/auth_test.go
```diff
@@ -880,6 +884,20 @@ func TestVerifyIDTokenAndCheckRevokedError(t *testing.T) {
 	}
 }
 
+func TestVerifyIDTokenAndCheckDisabledError(t *testing.T) {
```
hiranya911, line 887: In other languages we also added integration tests to verify this functionality. We might want to do the same here.

hiranya911 (changes requested):
Thanks for the changes. But the `checkRevokedOrDisabled()` helper function could use more work. We also need an integration test or two.

testdata/get_disabled_user.json
Galadros: Does it matter either way?  I simply copied get_user.json and modified it to be disabled.

hiranya911: Smaller, uncluttered resource files are easier to understand. Also prevents somebody from using them in test cases that were not originally intended.

auth/auth_test.go
Galadros: Where should I add integration tests?  I'm trying to find other examples of integration tests that use VerifyIDToken and/or checkRevokedOrDisabled, and having a bit of a hard time doing so.

integration/auth/auth_test.go
```diff
@@ -186,6 +186,39 @@ func TestRevokeRefreshTokens(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	if vt.UID != uid {
+		t.Errorf("UID = %q; want UID = %q", vt.UID, uid)
+	}
+
+	// Disable the user
+	updates := auth.UserToUpdate{}
+	updates.Disabled(true)
+	client.UpdateUser(context.Background(), uid, &updates)
```
hiranya911, line 212: Handle possible errors.

integration/auth/auth_test.go
```diff
@@ -186,6 +186,39 @@ func TestRevokeRefreshTokens(t *testing.T) {
...
+
+	vt, err = client.VerifyIDTokenAndCheckRevoked(context.Background(), idt)
+	we := "user has been disabled"
+	if vt != nil || err == nil || err.Error() != we {
```
hiranya911, line 216: Might also want to check `IsUserDisabled(err)` here.

auth/auth_test.go
```diff
@@ -880,6 +884,20 @@ func TestVerifyIDTokenAndCheckRevokedError(t *testing.T) {
...
+	s := echoServer(testGetDisabledUserResponse, t)
+	defer s.Close()
+	revokedToken := getIDToken(mockIDTokenPayload{"uid": "uid", "iat": 1970})
+	s.Client.idTokenVerifier = testIDTokenVerifier
+
+	p, err := s.Client.VerifyIDTokenAndCheckRevoked(context.Background(), revokedToken)
```
hiranya911, line 893: Just wondering. Do we have unit test coverage for `VerifySessionCookie()` with this change? We might want to add that too.

hiranya911 (changes requested):
Looks mostly good. Just a few minor things that can be added as improvements.

auth/auth_test.go
Galadros: The requested change was only to verifyIDToken, but regardless the two should function identically in unit tests (practically speaking, the existing unit test that was added is actually verifying that checkRevokedOrDisabled works properly.  verifyIDToken and VerifySessionCookie just directly return the error that they get from checkRevokedOrDisabled).

hiranya911: True. But `VerifySessionCookie()` is a separate public API with its own tests. See https://github.com/firebase/firebase-admin-go/blob/928b104e521ffc3d7de0319efe1553c31e85e13f/auth/auth_test.go#L1074

This PR is changing its behavior (rightly so), but without the proper test coverage we won't be able to detect future breakages. Imagine a scenario where we refactor how the `checkRevokedOrDisabled()` is used internally.  May be we decide to not call it from `VerifySessionCookie()`. The resulting behavior change will go unverified without a test case.

Also, for reference here's the equivalent PR in Java that we did: https://github.com/firebase/firebase-admin-java/pull/585/files. You will notice how there are separate tests for `VerifyIdToken()` and `VerifySessionCookie()`.

hiranya911 (approved):
LGTM with a comment about testing. Highly recommend adding the missing test case before merging.

auth/auth_test.go
bojeil-google: Hey @Galadros, @hiranya911 is right. I also recommend you add another test for `VerifySessionCookie()` for the exact same reason.

Galadros: Done.

PR #456, opened 2021-08-13
chore: Add nightly build workflow

lahirumaramba:
- Add nightly build workflow.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,0 +1,83 @@
+# Copyright 2021 Google Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Nightly Builds
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,83 @@
...
+        GOLINT=`go list -f {{.Target}} golang.org/x/lint/golint`
+        $GOLINT -set_exit_status ./...
+
+    - name: Run Tests
+      run: ./.github/scripts/run_all_tests.sh
+      env:
+        FIREBASE_SERVICE_ACCT_KEY: ${{ secrets.FIREBASE_SERVICE_ACCT_KEY }}
+        FIREBASE_API_KEY: ${{ secrets.FIREBASE_API_KEY }}
+
+    - name: Send email on failure
```
hiranya911, line 55: I've been thinking if we should try and reduce the duplication here. Something like:

```
- name: Generate email text
  if: failure() || cancelled()
  run: ./.github/scripts/generate_email.sh
  env:
    WORKFLOW_CANCELLED: ${{ cancelled() }}
    WORKFLOW_FAILED: ${{ failure() }}

- name: Send email
  uses: firebase/firebase-admin-node/.github/actions/send-email@master
  ...
```

No need to do this urgently. But something to think about.

hiranya911 (approved):
Thanks. LGTM with a suggestion for future improvement.

hiranya911:
I've changed base branch to dev.

.github/workflows/nightly.yml
lahirumaramba: Good call!
I will use one of the other repos to test this out first. Thanks!

lahirumaramba:
> I've changed base branch to dev.

Thank you!!

lahirumaramba:
@googlebot I fixed it.

PR #459, opened 2021-08-27
fix: Skipping credential lookup in emulator mode

### LINKED ISSUE #458: Authentication emulator fails on clean environment to required default credentials
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: WSL Unix
  * Firebase SDK version: 4.6.0
  * Library version: v4.6.0
  * Firebase Product: auth emulator

### [REQUIRED] Step 3: Describe the problem

Running auth emulator with the wanted env variable FIREBASE_AUTH_EMULATOR_HOST fails.

#### Steps to reproduce:

Start a project in a clean machine. Add env variable FIREBASE_AUTH_EMULATOR_HOST and make firebase auth app.
It gives error `google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.`

#### Relevant Code:

```go
conf := &firebase.Config{ProjectID: "demo-project"}
app, err := firebase.NewApp(ctx, conf)
auth, err := app.Auth(ctx)
if err != nil {
   log.Fatalf("making firebase auth failed: %v", err) // this line is hit
}
```

The problem was missing the option `firebase.NewApp(ctx, conf, option.WithoutAuthentication())`. This isn't documented anywhere. Took me an hour of searching google and trying all kinds of things.

I came across this as Github Action failed to it. Tested on a brand new EC2 ubuntu instance in Amazon and fails too. With `WithoutAuthentication()` option both work as expected.

hiranya911:
We currently perform a credential lookup in the emulator mode, which can result in errors if the developer hasn't specified any credentials during initialization. This PR attempts to solve this problem by defaulting to "owner" emulator credentials in the emulator mode. 

One downside of this solution is that it will cause all client options to get ignored in emulator mode. Ideally we should check if the developer has specified any credentials, and only replace that. But currently there's no clean way to implement such a replacement.

Also found a couple of existing unit tests that were taking 7 seconds each due to automatic retries. I've disabled retries for those tests thus saving 14 seconds from our unit test runs. 

Resolves #458

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -25,6 +25,8 @@ import (
 	"time"
 
 	"firebase.google.com/go/v4/internal"
+	"golang.org/x/oauth2"
+	"google.golang.org/api/option"
 	"google.golang.org/api/transport"
 )
 
@@ -45,6 +47,10 @@ const (
 var reservedClaims = []string{
 	"acr", "amr", "at_hash", "aud", "auth_time", "azp", "cnf", "c_hash",
 	"exp", "firebase", "iat", "iss", "jti", "nbf", "nonce", "sub",
+}
+
+var emulatorToken = &oauth2.Token{
+	AccessToken: "owner",
 }
 
 // Client is the interface for the Firebase auth service.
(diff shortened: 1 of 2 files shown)


auth/auth.go
```diff
@@ -114,7 +120,15 @@ func NewClient(ctx context.Context, conf *internal.AuthConfig) (*Client, error)
 		return nil, err
 	}
 
-	transport, _, err := transport.NewHTTPClient(ctx, conf.Opts...)
+	var opts []option.ClientOption
```
lahirumaramba, line 123: You must have tried this already... but just for my curiosity, can we access `conf.Opts` here without creating a new `[]option.ClientOption` and append the `ts`?

lahirumaramba (approved):
LGTM! Left one comment?

auth/auth.go
hiranya911: `conf` gets passed into this function as an argument. I'm not a huge fan of modifying function arguments, since the caller usually has a reference to the same struct.

yuchenshi (approved):

PR #467, opened 2021-11-05
fix(fcm): Fix typo in max number of tokens

lahirumaramba:
The API supports up to 500 tokens. Changing the comments in code to reflect that.
b/197867415


egilmorez (approved):
Thanks!

PR #468, opened 2021-11-15
feat(auth): enables OIDC Auth code flow

ScruffyProdigy:
Provides an option for developers to specify the OAuth response type for 
their OIDC provider (either one of these can be set:):

id_token
code (if set, must also set the client secret)

RELEASE NOTES: Added support for configuring the authorization code flow for OIDC providers.

### DIFF
--- a/auth/provider_config.go
+++ b/auth/provider_config.go
@@ -215,10 +215,8 @@ func (config *OIDCProviderConfigToCreate) buildRequest() (nestedMap, string, err
 		if val, ok := config.params.Get(idTokenResponseTypeKey); ok && val.(bool) {
 			return nil, "", errors.New("Only one response type may be chosen")
 		}
-	} else {
-		if val, ok := config.params.Get(idTokenResponseTypeKey); !ok || !val.(bool) {
-			return nil, "", errors.New("At least one response type must be returned")
-		}
+	} else if val, ok := config.params.Get(idTokenResponseTypeKey); !ok || !val.(bool) {
+		return nil, "", errors.New("At least one response type must be returned")
 	}
 
 	return config.params, config.id, nil
@@ -309,9 +307,7 @@ func (config *OIDCProviderConfigToUpdate) buildRequest() (nestedMap, error) {
 		if val, ok := config.params.Get(idTokenResponseTypeKey); ok && val.(bool) {
 			return nil, errors.New("Only one response type may be chosen")
 		}
-	}
-
-	if val, ok := config.params.Get(codeResponseTypeKey); ok && !val.(bool) {
+	} else if ok && !val.(bool) {
 		if val, ok := config.params.Get(idTokenResponseTypeKey); ok && !val.(bool) {
 			return nil, errors.New("At least one response type must be returned")
 		}
--- a/auth/provider_config.go
+++ b/auth/provider_config.go
@@ -38,11 +38,15 @@ const (
 	spEntityIDKey  = "spConfig.spEntityId"
 	callbackURIKey = "spConfig.callbackUri"
 
-	clientIDKey = "clientId"
-	issuerKey   = "issuer"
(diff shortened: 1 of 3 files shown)


auth/provider_config.go
```diff
@@ -180,6 +208,12 @@ func (config *OIDCProviderConfigToCreate) buildRequest() (nestedMap, string, err
 		return nil, "", fmt.Errorf("failed to parse Issuer: %v", err)
 	}
 
+	if val, ok := config.params.Get(codeResponseTypeKey); ok && val.(bool) {
+		if val, ok := config.params.GetString(clientSecretKey); !ok || val == "" {
+			return nil, "", errors.New("Client Secret must not be empty for Code Response Type")
```
lahirumaramba, line 213: Should we also check for cases where both is set to `true` and `false`? Also, where do we set `IDTokenResponseType` as the default?

auth/provider_config_test.go
```diff
@@ -303,6 +326,15 @@ func TestCreateOIDCProviderConfigInvalidInput(t *testing.T) {
 				ClientID("CLIENT_ID").
 				Issuer("not a url"),
 		},
+		{
+			name: "MissingClientSecret",
+			want: "Client Secret must not be empty for Code Response Type",
```
lahirumaramba, line 331: Should we add more test cases for when both are set to `true` and `false`?

lahirumaramba (changes requested):
Thanks, Ryan! Added a few comments.

auth/provider_config_test.go
ScruffyProdigy: Done

lahirumaramba (approved):
Thank you!
LGTM with minor comments. Let's get the reference docs reviewed as well.

auth/provider_config.go
```diff
@@ -113,11 +117,14 @@ func buildMask(data map[string]interface{}) []string {
...
-	ClientID    string
-	Issuer      string
+	ID                  string
+	DisplayName         string
+	Enabled             bool
+	ClientID            string
+	Issuer              string
+	ClientSecret        string
+	CodeResponseType    bool
+	IDTokenResponseType bool
```
kevinthecheung, line 127: `CodeResponseType` looks to me like it represents a type -- like a string or enum or something. Would `CodeResponseEnabled` etc be clearer?

egilmorez (approved):
LGTM

auth/provider_config.go
ScruffyProdigy: Sure.  Fixed

kevinthecheung (approved):

auth/provider_config.go
```diff
@@ -826,20 +901,30 @@ func (c *baseClient) makeRequest(
...
-	Issuer      string `json:"issuer"`
-	DisplayName string `json:"displayName"`
-	Enabled     bool   `json:"enabled"`
+	Name            string                   `json:"name"`
+	ClientID        string                   `json:"clientId"`
+	Issuer          string                   `json:"issuer"`
+	DisplayName     string                   `json:"displayName"`
+	Enabled         bool                     `json:"enabled"`
+	ClientSecret    string                   `json:"clientSecret"`
+	ResponseEnabled oidcProviderResponseType `json:"responseType"`
```
lahirumaramba, line 910: Nit: I believe this can stay as `ResponseType` since it is not a boolean type. Not a big deal though as it is internal implementation.

ScruffyProdigy: Oops, youre right.  Fixed

auth/provider_config.go
lahirumaramba: @ScruffyProdigy This changes the approved public API in go/gcip-oidc-provider-code-flow
Could you please check if this requires an amendment to the API proposal?

lahirumaramba (approved):

PR #469, opened 2021-11-25
add DEPRECATED labels on godoc

shogo82148:
godoc accepts “Deprecated:” comments as a depected feature.

https://go.dev/blog/godoc

> Sometimes a struct field, function, type, or even a whole package becomes redundant or unnecessary,
> but must be kept for compatibility with existing programs. To signal that an identifier should not be used,
> add a paragraph to its doc comment that begins with “Deprecated:” followed by some information about the deprecation.

For example, https://pkg.go.dev/net/http/httputil#ClientConn

image


lahirumaramba:
Duplicate of #531

PR #477, opened 2022-01-04
fix: Removing empty/unspecified fields from UserProvider

hiranya911:
Prevent serializing empty strings for unspecified fields in `auth.UserProvider`.


lahirumaramba (approved):
Thank you Hiranya! LGTM!

Issue #479, opened 2022-01-14
Update snippet for importing users with Firebase scrypt hashed passwords


thomasmburke:
### Describe the problem

When trying to migrate users across tenants in a test app I noticed that the `PasswordHash` of users I moved to the new tenant was being changed to the previous tenant's `PasswordHash` base64 encoded again. This means when a user tries to sign in with this newly imported user the correct password will be invalid

Additionally, when exporting users I noticed that some of the users' `PasswordHash` had `-` and `_` characters in it, so the previous function for decoding (`base64.StdEncoding.DecodeString()`) would throw an error when it encountered the byte containing that character. Instead we need to use [base64 URL encoding](https://pkg.go.dev/encoding/base64#URLEncoding) to decode, so we need a function that does just that (`base64.URLEncoding.DecodeString(s)`).

The hash key on the other hand can have a `/` character in it so it still requires [Standard base64 encoding](https://pkg.go.dev/encoding/base64#StdEncoding)

In short: the user's `PasswordHash` uses a different flavor of base 64 encoding then the signer key provided in the Firebase/Cloud Console and needs to be decoded properly before uploading a user.

PR #480, opened 2022-01-14
Updating user's PasswordHash to be base64 URL decoded before user import

thomasmburke:
Developers are having trouble migrating users because of the [issue described here](https://github.com/firebase/firebase-admin-go/issues/479). This update looks to update the snippet used in https://firebase.google.com/docs/auth/admin/import-users#import_users_with_firebase_scrypt_hashed_passwords to help users migrate users and avoid issues surrounding double base64 encoding.

### DIFF
--- a/snippets/auth.go
+++ b/snippets/auth.go
@@ -553,26 +553,35 @@ func importWithBcrypt(ctx context.Context, client *auth.Client) {
 
 func importWithScrypt(ctx context.Context, client *auth.Client) {
 	// [START import_with_scrypt]
-	users := []*auth.UserToImport{
-		(&auth.UserToImport{}).
-			UID("some-uid").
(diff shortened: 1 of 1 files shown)


thomasmburke:
So the update for this doc snippet updating two things:
1.) Decoding the user password hash
    - currently we don't so when a dev performs import the password hash is getting double encoding leading to invalid password errors when the correct password is provided for a newly imported user
2.) Updating the decoding functions so the hash.Scrypt key uses standard base64 encoding and user password hashes uses base64URL encoding
    - currently the snippet doesn't decode the use password hash, but it should using a base64URL encoding

kevinthecheung (approved):

lahirumaramba (approved):
Thank you @thomasmburke ! LGTM

PR #500, opened 2022-06-02
use go install to install golint

shogo82148:
use of go get to build and install packages is deprecated.

FYI, golint is also deprecated https://github.com/golang/go/issues/38968


lahirumaramba:
Thank you, @shogo82148 ! We have included some of these changes in the minimum Go version update #505 

Superseded by: #505

Issue #501, opened 2022-06-27
FR: Allow for sending of an out-of-band email action link


Dal-Papa:
### [READ] Step 1: Are you in the right place?

✅ 

### [REQUIRED] Step 2: Describe your environment

  * Operating System version: Google App Engine, Go 1.16 Runtime
  * Firebase SDK version: ?
  * Library version: v4.8.0
  * Firebase Product: auth

#### Steps to reproduce:

When using the `EmailSignInLink` or other functions that use `generateEmailActionLink` in the background, no email is sent despite the capability of the SDK to do so. 

#### Relevant Code:

```
link, err := s.Client.EmailSignInLink(context.Background(), testEmail, testActionCodeSettingsNoReturn)
if err != nil {
	t.Fatal(err)
}
```

PR #502, opened 2022-06-27
feat(auth): Add returnOobLink to the ActionCodeSettings

Dal-Papa:
### Discussion

https://github.com/firebase/firebase-admin-go/issues/501

### Testing

  * Make sure all existing tests in the repository pass after your change. ✅ 
  * If you fixed a bug or added a feature, add a new test to cover your code. ✅ 

### API Changes

    can discuss it together.

### DIFF
--- a/auth/email_action_links_test.go
+++ b/auth/email_action_links_test.go
@@ -99,9 +99,8 @@ func TestEmailVerificationLink(t *testing.T) {
 	}
 
 	want := map[string]interface{}{
-		"requestType":   "VERIFY_EMAIL",
-		"email":         testEmail,
-		"returnOobLink": true,
+		"requestType": "VERIFY_EMAIL",
+		"email":       testEmail,
 	}
 	if err := checkActionLinkRequest(want, s); err != nil {
 		t.Fatalf("EmailVerificationLink() %v", err)
@@ -154,9 +153,8 @@ func TestPasswordResetLink(t *testing.T) {
 	}
 
 	want := map[string]interface{}{
-		"requestType":   "PASSWORD_RESET",
-		"email":         testEmail,
-		"returnOobLink": true,
+		"requestType": "PASSWORD_RESET",
+		"email":       testEmail,
 	}
 	if err := checkActionLinkRequest(want, s); err != nil {
 		t.Fatalf("PasswordResetLink() %v", err)
--- a/auth/tenant_mgt_test.go
+++ b/auth/tenant_mgt_test.go
@@ -503,9 +503,8 @@ func TestTenantEmailVerificationLink(t *testing.T) {
 	}
 
 	want := map[string]interface{}{
-		"requestType":   "VERIFY_EMAIL",
-		"email":         testEmail,
-		"returnOobLink": true,
+		"requestType": "VERIFY_EMAIL",
+		"email":       testEmail,
 	}
 	if err := checkActionLinkRequestWithURL(want, wantEmailActionURL, s); err != nil {
 		t.Fatalf("EmailVerificationLink() %v", err)
@@ -530,9 +529,8 @@ func TestTenantPasswordResetLink(t *testing.T) {
 	}
 
 	want := map[string]interface{}{
-		"requestType":   "PASSWORD_RESET",
-		"email":         testEmail,
-		"returnOobLink": true,
+		"requestType": "PASSWORD_RESET",
+		"email":       testEmail,
 	}
 	if err := checkActionLinkRequestWithURL(want, wantEmailActionURL, s); err != nil {
 		t.Fatalf("PasswordResetLink() %v", err)


Dal-Papa:
@lahirumaramba @hiranya911 👋

lahirumaramba:
Hi @Dal-Papa Thank you for your patience on this. The Admin SDKs currently do not support sending email action links. Setting `returnOobLink` does not send the email, in-fact we already set it to true in the code below. https://github.com/firebase/firebase-admin-go/blob/23a1f17959238d989add41d562602b8bf04e71d6/auth/email_action_links.go#L119

If you are looking to generate email action links the SDK already supports that and you can use your own email API to send out the emails.

Dal-Papa:
This feature use to work as we were able to send an email through Firebase to our users for them to recovery their account, what has changed ? Also, what's the difference between this feature and the SignIn With Email Link where you are sending the email to the user ? 
We do not want to manage an email API as we're using Firebase already to deal with that.

Dal-Papa:
@lahirumaramba : Either you read the code wrong or you used it wrong but I was able to make Firebase send me an email with my version of the code. 
https://github.com/firebase/firebase-admin-go/blob/23a1f17959238d989add41d562602b8bf04e71d6/auth/email_action_links.go#L119
This is overwritten by the settings object so therefore it becomes false and hence it sends an email. Please consider this PR, we are using the fork in the meantime.

prameshj:
> Lahiru Maramba : Either you read the code wrong or you used it wrong but I was able to make Firebase send me an email with my version of the code.
> 
> https://github.com/firebase/firebase-admin-go/blob/23a1f17959238d989add41d562602b8bf04e71d6/auth/email_action_links.go#L119
> 
> 
> This is overwritten by the settings object so therefore it becomes false and hence it sends an email. Please consider this PR, we are using the fork in the meantime.

Hi @Dal-Papa , just to clarify, you are exposing ReturnOObLink in ActionCodeSettings, so you can override to false, in case you want the email to be sent automatically, correct? That makes sense, however, I am curious about the use-case.. typically the email verification/password reset etc are triggered by the client, so maybe more suitable in the client SDKs? Can you explain how it is useful in admin SDKs? Happy to move forward on this, but would like to better understand the scenario. Thanks!

auth/email_action_links_test.go
```diff
@@ -40,6 +40,7 @@ var testActionCodeSettings = &ActionCodeSettings{
 	AndroidPackageName:    "com.example.android",
 	AndroidInstallApp:     true,
 	AndroidMinimumVersion: "6",
+	ReturnOobLink:         true,
```
prameshj, line 43: Can we remove this in the base testActionCodeSettings struct? We need a test case to verify that if this field is not specified, returnOobLink is still set to true, due to sdk implementation.

I suggest adding a new test case in  the newly added test, see my other comment. Thanks!

auth/email_action_links_test.go
```diff
@@ -309,6 +311,35 @@ func TestEmailVerificationLinkError(t *testing.T) {
 	}
 }
 
+func TestEmailVerificationSendEmail(t *testing.T) {
+	s := echoServer(testActionLinkResponse, t)
+	defer s.Close()
+
```
prameshj, line 317: Can we call this "TestEmailVerificationCustomOobLinkSettings" or similar?

Then, we can add 2 cases:
```
cases := []bool{true, false}

testActionCodeSettingsCustom := map[string]interface{}{}
// make a copy to avoid modifying the base maps.
for k, v := range testActionCodeSettings {
   testActionCodeSettingsCustom[k] = v 
}
testActionCodeSettingsMapCustom :=  map[string]interface{}{}
for k, v := range testActionCodeSettingsMapCustom {
   testActionCodeSettingsMapCustom[k] = v 
}

for _, returnOobLink := range cases {
  testActionCodeSettingsCustom.ReturnOobLink = returnOobLink
  testActionCodeSettingsCustomMap["returnOobLink"] = returnOobLink
 link, err := s.Client.EmailVerificationLink(..., testActionCodeSettingsCustom)
...
...
}
```

It would be great if you can add a similar test for EmailSignInLink and PasswordReset too. Thanks!

prameshj:
> > > Lahiru Maramba : Either you read the code wrong or you used it wrong but I was able to make Firebase send me an email with my version of the code.
> > > https://github.com/firebase/firebase-admin-go/blob/23a1f17959238d989add41d562602b8bf04e71d6/auth/email_action_links.go#L119
> > > 
> > > This is overwritten by the settings object so therefore it becomes false and hence it sends an email. Please consider this PR, we are using the fork in the meantime.
> > 
> > 
> > Hi Clement DAL PALU , just to clarify, you are exposing ReturnOObLink in ActionCodeSettings, so you can override to false, in case you want the email to be sent automatically, correct? That makes sense, however, I am curious about the use-case.. typically the email verification/password reset etc are triggered by the client, so maybe more suitable in the client SDKs? Can you explain how it is useful in admin SDKs? Happy to move forward on this, but would like to better understand the scenario. Thanks!
> 
> That's correct ! We do this so we can make a few checks on the account in the backend before sending such link (i.e. if the account isn't banned or if their email has been verified in the past).

Thanks! Left a couple of comments in the unit test, LGTM otherwise.

lahirumaramba:
@Dal-Papa Thank you for your contribution! I definitely did misunderstood the API in my initial comment, my apologies!

@prameshj I think we need an internal API proposal for the public API change before we can merge this. Thanks!

prameshj (approved):
Thanks! We are working on the API review doc.

auth/email_action_links.go
```diff
@@ -32,6 +32,7 @@ type ActionCodeSettings struct {
 	AndroidMinimumVersion string `json:"androidMinimumVersion,omitempty"`
 	AndroidInstallApp     bool   `json:"androidInstallApp,omitempty"`
 	DynamicLinkDomain     string `json:"dynamicLinkDomain,omitempty"`
+	ReturnOobLink         bool   `json:"returnOobLink"`
```
pragatimodi, line 35: The naming for the field is a bit weird. Since we want to keep the default value of the boolean as false, it would be better to rename this variable to `SendEmailLink` instead.

pragatimodi, line 35: API Review feedback was to rename this variable to `SendEmailLink` in order to maintain the default value false default value for boolean

pragatimodi: @Dal-Papa Would you be able to make this change, please?

Dal-Papa: As you want but that will break the override that was happening on the key name due to the `toMap` function...

auth/email_action_links_test.go
```diff
@@ -111,24 +111,30 @@ func TestEmailVerificationLinkWithSettings(t *testing.T) {
...
 	defer s.Close()
 
-	link, err := s.Client.EmailVerificationLinkWithSettings(context.Background(), testEmail, testActionCodeSettings)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if link != testActionLink {
-		t.Errorf("EmailVerificationLinkWithSettings() = %q; want = %q", link, testActionLink)
-	}
+	cases := []bool{true, false}
```
prameshj, line 114: I'm wondering if we can make this a list of testActionCodeSettings and testActionCodeSettingsMap instead, so we also test the "unset" case. Added some sample here, to illustrate what i mean.. (I did not verify that this compiles)

```
sendEmailLinkValues := []string{"true", "false", "unset"}
testActionCodeSettingsCases := []ActionCodeSettings{}
testActionCodeSettingsMapCases := []map[string]interface{}{}
for _, val := range sendEmailLinkValues {
  settings, settingsMap:= getCopiesOfTestSettings(testActionCodeSettings,
		testActionCodeSettingsMap)
  switch(val) {
  case "true":
    settings.SendEmailLink = true;
    settingsMap["returnOobLink"] = false;
  case "false":
     settings.SendEmailLink = false;
     settingsMap["returnOobLink"] = true;
  case "unset":
  }
  testActionCodeSettingsCases = append(testActionCodeSettingsCases, settings)
 testActionCodeSettingsMapCases = append(testActionCodeSettingsMapCases, settingsMap)
}
```

auth/email_action_links.go
```diff
@@ -58,6 +59,7 @@ func (settings *ActionCodeSettings) toMap() (map[string]interface{}, error) {
 	if err := json.Unmarshal(b, &result); err != nil {
 		return nil, err
 	}
+	result["returnOobLink"] = !settings.SendEmailLink
```
prameshj, line 62: Maybe add a comment here like : "Since SendEmailLink defaults to true, we will always set "returnOobLink" to false by default, this means no email is sent out in the defautl case".

And we can remove the `"returnOobLink" : true` setting in line 121?

prameshj (commented):
Thanks for making the changes! Left 2 more comments (sorry for the back and forth). Other than that, LGTM to me.

Dal-Papa:
Hey @prameshj & @lahirumaramba ! I've fixed your comments and I've recently merged the latest changes, would you mind taking a look so we can merge this old PR ? 🙂

auth/email_action_links.go
Dal-Papa: Fixed

auth/email_action_links_test.go
Dal-Papa: Done

auth/email_action_links.go
prameshj: Looking at this after several weeks - can you remind me how SendEmailLink defaults to true? I think the comment should be:

"Since SendEmailLink defaults to false, we will always set "returnOobLink" to true by default, this means no email is sent out in the default case".

Apologies for the mistake in my earlier comment.

prameshj (approved):
lgtm, except one comment about fixing a code comment. Thanks for your patience!

drmaples:
any updates on when this will be merged?

Dal-Papa:
@prameshj : I've updated to the latest upstream branch and fixed the comment. Please take a look.

Dal-Papa:
@lahirumaramba : It looks like @prameshj is AWOL, can you please take over ?

Dal-Papa:
@jonathanedey : You seem like the only person still active here, can you please take a look ?

PR #511, opened 2022-09-14
feat(auth): MFA Support for GoLang createUser, updateUser

pragatimodi:


auth/user_mgt.go
```diff
@@ -604,6 +649,41 @@ func validateProvider(providerID string, providerUID string) error {
...
 }
 
+func validateAndFormatMfaSettings(mfaSettings MultiFactorSettings, methodType string) ([]*multiFactorInfoResponse, error) {
+	var mfaInfo []*multiFactorInfoResponse
+	if len(mfaSettings.EnrolledFactors) != 0 {
+		for _, multiFactorInfo := range mfaSettings.EnrolledFactors {
+			if multiFactorInfo.FactorID == "" {
+				return nil, fmt.Errorf("no factor id specified")
+			}
+			if methodType == createUserMethod {
```
prameshj, line 659: nit: how about we make this a switch statement?

```
switch methodType {
case createUserMethod:
  // enrollment timestamp and uid should be empty
case updateUserMethod:
// uid should be specified.
default:
// unsupported method.
}

// Phone number, display name validations.
```

auth/user_mgt_test.go
```diff
@@ -713,6 +769,23 @@ var createUserCases = []struct {
 	{
 		(&UserToCreate{}).PhotoURL("http://some.url"),
 		map[string]interface{}{"photoUrl": "http://some.url"},
+	}, {
```
prameshj, line 772: nit: maybe one more test case with multiple mfa phone numbers?

auth/user_mgt_test.go
```diff
@@ -772,6 +845,41 @@ func TestInvalidUpdateUser(t *testing.T) {
 		}, {
 			(&UserToUpdate{}).Password("short"),
 			"password must be a string at least 6 characters long",
+		}, {
+			(&UserToUpdate{}).MFASettings(MultiFactorSettings{
+				EnrolledFactors: []*MultiFactorInfo{
+					{
+						UID:         "enrolledSecondFactor1",
+						PhoneNumber: "+16505557348",
+						DisplayName: "",
```
prameshj, line 854: You can drop this field, it will be "" by default.

auth/user_mgt_test.go
```diff
@@ -2051,9 +2195,10 @@ type mockAuthServer struct {
 // echoServer takes either a []byte or a string filename, or an object.
 //
 // echoServer returns a server whose client will reply with depending on the input type:
-//   * []byte: the []byte it got
-//   * object: the marshalled object, in []byte form
-//   * nil: "{}" empty json, in case we aren't interested in the returned value, just the marshalled request
+//   - []byte: the []byte it got
```
prameshj, line 2198: I am curious why we need this change. Are rest of the comment blocks using - instead of *?

pragatimodi: Looks like it was auto generated. Reverting it to see if checks are still passing.

pragatimodi: Looks like the linter on my editor just saves like this. Doesn't make a difference to the code as there are no other itemized comment blocks.

auth/user_mgt.go
```diff
@@ -604,6 +649,41 @@ func validateProvider(providerID string, providerUID string) error {
...
+				if multiFactorInfo.EnrollmentTimestamp != 0 {
+					return nil, fmt.Errorf("\"EnrollmentTimeStamp\" is not supported when adding second factors via \"createUser()\"")
```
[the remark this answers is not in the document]
pragatimodi: Enrollment time and uid are not allowed for signupNewUser endpoint. They will automatically be provisioned server side.

auth/user_mgt.go
```diff
@@ -604,6 +649,45 @@ func validateProvider(providerID string, providerUID string) error {
...
+			// Enrollment time and uid are not allowed for signupNewUser endpoint. They will automatically be provisioned server side.
+			if multiFactorInfo.EnrollmentTimestamp != 0 {
+				return nil, fmt.Errorf("\"EnrollmentTimeStamp\" is not supported when adding second factors via \"createUser()\"")
+			}
+			if multiFactorInfo.UID != "" {
+				return nil, fmt.Errorf("\"uid\" is not supported when adding second factors via \"createUser()\"")
+			}
+		case updateUserMethod:
+			if multiFactorInfo.UID == "" {
+				return nil, fmt.Errorf("the second factor \"uid\" must be a valid non-empty string")
```
prameshj, line 669: nit: include "must be a valid non-empty string when adding second factors via \"updateUser()\"" like you did for createUser.

auth/user_mgt.go
```diff
@@ -604,6 +649,45 @@ func validateProvider(providerID string, providerUID string) error {
...
+			}
+		default:
+			return nil, fmt.Errorf("unsupported method")
```
prameshj, line 672: include "methodType" in the error.

auth/user_mgt.go
```diff
@@ -604,6 +649,45 @@ func validateProvider(providerID string, providerUID string) error {
...
+		}
+		if multiFactorInfo.FactorID == phoneMultiFactor {
+			if err := validatePhone(multiFactorInfo.PhoneNumber); err != nil {
+				return nil, fmt.Errorf("the second factor \"phoneNumber\" for \"%s\" must be a non-empty E.164 standard compliant identifier string", multiFactorInfo.PhoneNumber)
+			}
+			if err := validateDisplayName(multiFactorInfo.DisplayName); err != nil {
+				return nil, fmt.Errorf("the second factor \"displayName\" for \"%s\" must be a valid non-empty string", multiFactorInfo.PhoneNumber)
```
prameshj, line 679: use multiFactor.DisplayName instead of PhoneNumber in the error string.

auth/user_mgt.go
```diff
@@ -604,6 +649,45 @@ func validateProvider(providerID string, providerUID string) error {
...
+				return nil, fmt.Errorf("the second factor \"uid\" must be a valid non-empty string")
+			}
+		default:
+			return nil, fmt.Errorf("unsupported method")
+		}
+		if multiFactorInfo.FactorID == phoneMultiFactor {
+			if err := validatePhone(multiFactorInfo.PhoneNumber); err != nil {
+				return nil, fmt.Errorf("the second factor \"phoneNumber\" for \"%s\" must be a non-empty E.164 standard compliant identifier string", multiFactorInfo.PhoneNumber)
+			}
+			if err := validateDisplayName(multiFactorInfo.DisplayName); err != nil {
```
prameshj, line 678: This can be outside the 'if multiFactorInfo.FactorID == phoneMultiFactor {' check.

prameshj (commented):
Please add integration tests for create and update user. Once that's done and the comments here are addressed, feel free to get review from @lahirumaramba and submit. Thanks!

auth/user_mgt_test.go
```diff
@@ -642,6 +642,62 @@ func TestInvalidCreateUser(t *testing.T) {
 		}, {
 			(&UserToCreate{}).Email("a@a@a"),
 			`malformed email string: "a@a@a"`,
+		}, {
+			(&UserToCreate{}).MFASettings(MultiFactorSettings{
+				EnrolledFactors: []*MultiFactorInfo{
+					{
+						UID:         "EnrollmentID",
+						PhoneNumber: "+16505557348",
```
renkelvin, line 650: Let's use a fake number for tests here and elsewhere.

auth/user_mgt.go
```diff
@@ -39,6 +39,9 @@ const (
 
 	// Maximum number of users allowed to batch delete at a time.
 	maxDeleteAccountsBatchSize = 1000
+	createUserMethod           = "createUser"
+	updateUserMethod           = "updateUser"
+	phoneMultiFactor           = "phone"
```
renkelvin, line 44: Maybe "phoneMultiFactorID"

renkelvin (approved):

auth/user_mgt.go
```diff
@@ -66,6 +69,7 @@ type multiFactorInfoResponse struct {
 }
 
 // MultiFactorInfo describes a user enrolled second phone factor.
+// TODO : convert PhoneNumber to PhoneMultiFactorInfo struct
```
lahirumaramba, line 72: Is this still a TODO, I wonder if this will show up in the reference docs for `MultiFactorInfo` type.

auth/user_mgt.go
```diff
@@ -155,10 +164,34 @@ func (u *UserToCreate) set(key string, value interface{}) *UserToCreate {
 	return u
 }
 
+// Converts a client format second factor object to server format.
+func convertMultiFactorInfoToServerFormat(mfaInfo MultiFactorInfo) (multiFactorInfoResponse, error) {
+	var authFactorInfo multiFactorInfoResponse
+	if mfaInfo.EnrollmentTimestamp != 0 {
+		authFactorInfo.EnrolledAt = time.Unix(mfaInfo.EnrollmentTimestamp, 0).Format("2006-01-02T15:04:05Z07:00Z")
```
lahirumaramba, line 171: `2006-01-02T15:04:05Z07:00Z` Do we need all 9 fractions here? Example in the messaging API: https://github.com/firebase/firebase-admin-go/blob/23a1f17959238d989add41d562602b8bf04e71d6/messaging/messaging.go#L51

lahirumaramba: Nit: Could we also make the format a const?

lahirumaramba (approved):
Thank you! LGTM!

lahirumaramba:
Adding @kevinthecheung to review the reference doc changes. Thank you!

auth/user_mgt.go
pragatimodi: This is a TODO for later stages when we want to add more MFA methods such as TOTP, etc. If it does show up in the reference docs, we can report this as a bug and remove from here.

kevinthecheung (approved):

brokeyourbike:
Hi @lahirumaramba @kevinthecheung , when do you think this can be merged?

auth/user_mgt.go
pragatimodi: Will accommodate these changes in a separate commit. Okay to keep the granularity of the 9 fractions for this.

PR #530, opened 2022-12-13
fix(auth): Allow to update MFA 

### LINKED ISSUE #529: BUG: UpdateUser() is not able to update MFA
### Describe your environment

  * Operating System version: Mac OS Ventura 13.0.1
  * Firebase SDK version: Firebase Admin SDK v4
  * Library version: v4.10.0
  * Firebase Product: auth

### Describe the problem

It's not possible to update multi factor enrollments with `UpdateUser()` function. Create and update API requests have different structure when we speaking about MFA (don't know who thought it will be cool to do that). It's also not required to pass fields like `UID` or `DisplayName` when updating or creating MFA.

We can see the valid implementation in the [firebase-admin-node repository](https://github.com/firebase/firebase-admin-node/blob/b2a28aee5cac172ed183bc47f71a55b31f79696a/test/unit/auth/auth-api-request.spec.ts#L2034)

#### Steps to reproduce:

1. Create user (or have existing firebase user) without MFA
2. Try to add MFA with `UpdateUser`
3. Nothing will happen, used will not have MFA after update

#### Relevant Code:

```go
// 1. Create user
createParams := (&auth.UserToCreate{}).
		UID("01GM6YCBT7T0VNPNTCP44S7E8G").
		Email("john@doe.com").
		EmailVerified(true).
		Password("secret").
		DisplayName("John Doe").
		Disabled(false).
                MFASettings(auth.MultiFactorSettings{EnrolledFactors: nil})

user, err := firebaseAuth.CreateUser(ctx, params)
require.NoError(t, err)

// 2. Update user
updateParams := (&auth.UserToUpdate{}).
		MFASettings(auth.MultiFactorSettings{
			EnrolledFactors: []*auth.MultiFactorInfo{{
                                // problems will arise even on this stage
                                // because library things that it's required to pass UID and DisplayName
				FactorID:    "phone",
				PhoneNumber: "+16505557348",
			}},
		})

user, err = firebaseAuth.UpdateUser(ctx, user.UID, params)
assert.NoError(t, err)
assert.Len(t, user.MultiFactor.EnrolledFactors, 1)
```

brokeyourbike:
- `UpdateUser` can update MFA
- `UID` is not required when updating MFA
- `DisplayName` is not required field

Fixes #529.

RELEASE NOTE: Allow updating multi factor enrollments in `UpdateUser()` API


brokeyourbike:
@lahirumaramba @hiranya911 👋

prameshj:
Thanks for the fix! 
@pragatimodi can you take a look too? This is a follow up to https://github.com/firebase/firebase-admin-go/pull/511, the fields in updateUser expect "mfa" rather than "mfaInfo". Thanks!

auth/user_mgt.go
```diff
@@ -170,14 +174,16 @@ func convertMultiFactorInfoToServerFormat(mfaInfo MultiFactorInfo) (multiFactorI
 	if mfaInfo.EnrollmentTimestamp != 0 {
 		authFactorInfo.EnrolledAt = time.Unix(mfaInfo.EnrollmentTimestamp, 0).Format("2006-01-02T15:04:05Z07:00Z")
 	}
+	if mfaInfo.UID != "" {
```
prameshj, line 177: Since phone is the only factorId, is it ok to put this in the if block below?

Or is this to cover the case where a factorId is not specified?

auth/user_mgt.go
```diff
@@ -333,7 +339,7 @@ func (u *UserToUpdate) validatedRequest() (map[string]interface{}, error) {
 			if err != nil {
 				return nil, err
 			}
-			req["mfaInfo"] = mfaInfo
+			req["mfa"] = multiFactorEnrollments{mfaInfo}
```
prameshj, line 342: nit: Can you link to the docs https://cloud.google.com/identity-platform/docs/reference/rest/v1/accounts/update for easy reference next time? Thanks!

auth/user_mgt.go
```diff
@@ -665,15 +671,9 @@ func validateAndFormatMfaSettings(mfaSettings MultiFactorSettings, methodType st
 				return nil, fmt.Errorf("\"uid\" is not supported when adding second factors via \"createUser()\"")
 			}
 		case updateUserMethod:
-			if multiFactorInfo.UID == "" {
-				return nil, fmt.Errorf("the second factor \"uid\" must be a valid non-empty string when adding second factors via \"updateUser()\"")
-			}
 		default:
 			return nil, fmt.Errorf("unsupported methodType: %s", methodType)
 		}
-		if err := validateDisplayName(multiFactorInfo.DisplayName); err != nil {
```
prameshj, line 674: Wouldn't we want to include a display name for a (new or existing) mfa option? Any downside to requiring this?

auth/user_mgt_test.go
```diff
@@ -894,17 +873,6 @@ func TestInvalidUpdateUser(t *testing.T) {
 				},
 			}),
 			`the second factor "phoneNumber" for "invalid" must be a non-empty E.164 standard compliant identifier string`,
-		}, {
```
prameshj, line 897: nit: I would prefer to keep the test cases related to display name field being required.

auth/user_mgt.go
brokeyourbike: It can be moved to the block bellow of course, but why we have the `EnrollmentTimestamp` at the top for example?

I just thought that `UID` will probably exist for every "factor", because we will need it when updating something, so it's safe to move it from the block

auth/user_mgt.go
brokeyourbike: the only downside it that it's not really required from the API point of view

auth/user_mgt.go
prameshj: Good point. We will likely move it out of the if block when we add another MFA (like TOTP).

Similarly, DisplayName is a generic field too. So you could move that out too.

auth/user_mgt.go
prameshj: I think it would be ok to require it on the SDK, since this field helps identify the MFA option easily. @lahirumaramba any thoughts?

pragatimodi (approved):

pragatimodi:
Thanks for finding this fix!

auth/user_mgt.go
prameshj: Super nit: Please add this link as a code comment if possible.

prameshj:
Thanks!  I have one final nit, LGTM otherwise.

@lahirumaramba for final approval/merge.

brokeyourbike:
can this be merged @lahirumaramba ?

auth/user_mgt.go
prameshj: This change is not showing up in the diff yet.. otherwise lgtm

brokeyourbike: Have added the link as you suggested @prameshj . Please take a look

brokeyourbike:
Hi team, can you please take a look at the PR. I have made the changes as requested. @prameshj @lahirumaramba

prameshj (approved):

brokeyourbike:
Hello team, can we please move forward with this PR? Let me know if there is anything I should do
@prameshj @pragatimodi @lahirumaramba

prameshj:
Thanks for checking, Ivan! The changes look good to me. I think we only need @lahirumaramba to approve. Lahiru, can you help with this? Thanks!

brokeyourbike:
Are my proposed changes outdated and do not comply with expected release? Should I create a new PR and revisit what was done?

Please give at least some direction @prameshj @pragatimodi @lahirumaramba , thanks.

lahirumaramba:
Thanks folks! We will include this in the next release.

PR #535, opened 2023-02-01
MFA Config with TOTP Provider

pragatimodi:


auth/config.go
```diff
@@ -0,0 +1,172 @@
...
+	constMfaObject                = "MultiFactorConfig"
+	constProviderConfigObject     = "ProviderConfig"
+	constTotpProviderConfigObject = "TotpProviderConfig"
+	constState                    = "state"
+	constEnabledProviders         = "enabledProviders"
+	constFactorIds                = "factorIds"
+)
+
+var validAuthFactors = map[string]bool{
+	"PHONE_SMS": true,
```
renkelvin, line 24: Should this include SMS or TOTP?

pragatimodi: SMS only, we've switch to `providerConfigs` for factors other than SMS.

renkelvin (approved):
LGTM, thanks!

auth/config.go
```diff
@@ -0,0 +1,172 @@
...
+func validateConfigKeys(inputReq *map[string]interface{}, validKeys map[string]bool, configName string) error {
+	for key := range *inputReq {
+		if !validKeys[key] {
+			return fmt.Errorf(`"%s" is not a valid %s parameter`, key, configName)
+		}
+	}
+	return nil
+}
+
+func stringKeys(m *map[string]bool) string {
```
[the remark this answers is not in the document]
pragatimodi: I meant string as a verb but I get that it could be confusing.
Thanks for pointing this out!

PR #531, opened 2022-12-16
fix: Use the correct deprecated syntax in code

shogo82148:
The godoc of deprecated types, values and functions should be match the regexp `(^|\n\s*\n)\s*Deprecated:`.

https://github.com/golang/pkgsite/blob/8996ff632abee854aef1b764ca0501f262f8f523/internal/godoc/dochtml/deprecated.go#L12-L13

pkg.go.dev marks these items deprecated and show them using a special style.

e.g. https://pkg.go.dev/net/http#CloseNotifier

image

PR #544, opened 2023-03-21
Implement `SendEach`, `SendEachDryRun`, `SendEachForMulticast`,

Doris-Ge:
`SendEachForMulticastDryRun`

1. Add `SendEach`, `SendEachDryRun`, `SendEachForMulticast`, `SendEachForMulticastDryRun`
2. Deprecate `SendAll`, `SendAllDryRun`, `SendMulticast`, `SendMulticastDryRun`

`SendEach` vs `SendAll`
1. `SendEach` sends one HTTP request to V1 Send endpoint for each message in the array. `SendAll` sends only one HTTP request to V1 Batch Send endpoint to send all messages in the array.
2. `SendEach` calls fcmClient.Send to send each message and constructs a SendResponse with the returned message id or error. `SendEach` uses sync.WaitGroup to execute all fcmClient.Send calls asynchronously and wait for all of them to complete and construct a BatchResponse with all SendResponses. Therefore, unlike `SendAll`, `SendEach` does not always returns an error for a total failure. It can also return a `BatchResponse` with only errors in it.

`SendEachForMulticast` calls `SendEach` under the hood.

Will send the integration tests in another PR.

### DIFF
--- a/messaging/messaging_batch.go
+++ b/messaging/messaging_batch.go
@@ -27,6 +27,7 @@ import (
 	"mime/multipart"
 	"net/http"
 	"net/textproto"
+	"sync"
 
 	"firebase.google.com/go/v4/internal"
 )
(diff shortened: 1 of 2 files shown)


messaging/messaging_batch.go
```diff
@@ -122,6 +249,8 @@ func (c *fcmClient) SendAllDryRun(ctx context.Context, messages []*Message) (*Ba
 // responses list obtained from the return value corresponds to the order of the input tokens. An
 // error from SendMulticast indicates a total failure -- i.e. the message could not be sent to any
 // of the recipients. Partial failures are indicated by a `BatchResponse` return value.
+//
+// Deprecated. Use SendEachForMulticast instead.
```
lahirumaramba, line 253: I think this should be `Deprecated: Use SendEachForMulticast instead.`
See: https://go.dev/blog/godoc

messaging/messaging_batch.go
```diff
@@ -142,6 +271,8 @@ func (c *fcmClient) SendMulticast(ctx context.Context, message *MulticastMessage
 // the return value corresponds to the order of the input tokens. An error from SendMulticastDryRun
 // indicates a total failure -- i.e. none of the messages were sent to FCM for validation. Partial
 // failures are indicated by a `BatchResponse` return value.
+//
+// Deprecated. Use SendEachForMulticastDryRun instead.
```
lahirumaramba, line 275:
```suggestion
// Deprecated: Use SendEachForMulticastDryRun instead.
```

lahirumaramba (approved):
Thanks @Doris-Ge ! Looks great.

lahirumaramba:
Please make sure that this branch is based off of `dev` and not `mater`. In the Go repo, we make all the new changes in the `dev` branch and only the releases are merged to the main (`master`) branch. Thank you!

messaging/messaging_batch.go
lahirumaramba: Good catch! The current usage of `Deprecated` in the SDK is actually incorrect. There is a fix in #531

PR #552, opened 2023-04-05
feat(fcm): SendEach and SendEachForMulticast for FCM batch send

Doris-Ge:
- Deprecate `sendAll()` and `sendMulticast()` APIs.
- Add new `sendEach()` and `sendEachForMulticast()` APIs.

### DIFF
--- a/messaging/messaging_batch.go
+++ b/messaging/messaging_batch.go
@@ -93,7 +93,7 @@ type BatchResponse struct {
 // The messages array may contain up to 500 messages. Unlike `SendAll()`, SendEach sends the entire
 // array of messages by making a single HTTP call for each message. The responses list
 // obtained from the return value corresponds to the order of the input messages. An error
-// from SendEach or a `BatchResponse` with all failures indicates a total failure -- i.e.
+// from SendEach or a `BatchResponse` with all failures indicates a total failure, meaning that
 // none of the messages in the list could be sent. Partial failures or no failures are only
 // indicated by a `BatchResponse` return value.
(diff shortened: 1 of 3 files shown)


lahirumaramba:
Thank you @Doris-Ge. Please get the reference docs reviewed by @egilmorez before we merge this.

messaging/messaging_batch.go
```diff
@@ -87,14 +88,138 @@ type BatchResponse struct {
 	Responses    []*SendResponse
 }
 
+// SendEach sends the messages in the given array via Firebase Cloud Messaging.
```
egilmorez, line 91: Is this the comment that will be parsed into reference docs?

Normally, I would be asking for all literals to be backticked; but since that does not at all seem to be the standard in pages like https://pkg.go.dev/firebase.google.com/go@v3.13.0+incompatible/messaging#SendResponse I'll suggest the opposite, so our approach is more or less consistent.

So, I think things like SendAll()` and `BatchResponse` should actually *remove* the backticks.

In any case, we don't want a mix, like "An error
// from SendEach or a `BatchResponse` ...."

Any thoughts here Lahiru?  Thanks!

lahirumaramba: Yep, these comments will appear in the reference docs. Go reference docs do not render backticks, but we have used them in some places in the past... I am good with us removing them here. Thanks, Eric!

Reference: https://pkg.go.dev/firebase.google.com/go/v4/messaging#Client.Send

lahirumaramba (approved):

PR #569, opened 2023-07-17
Add MultiplefactorConfig.state support in Project and Tenant config

ynaka81:
- Add `MultiplefactorConfig.state` support in Project and Tenant config which has been implemented in [firebase-admin-node](https://github.com/firebase/firebase-admin-node).
- When I asked Firebase support, I was sure that you were planning to support this feature in the future. That's why I directly created a PR but if you think it's better to create a new issue first, I will follow the convention in this repository.

### DIFF
--- a/auth/multi_factor_config_mgt.go
+++ b/auth/multi_factor_config_mgt.go
@@ -45,6 +45,8 @@ const (
 // MultiFactorConfig represents a multi-factor configuration for a tenant or project.
 // This can be used to define whether multi-factor authentication is enabled or disabled and the list of second factor challenges that are supported.
 type MultiFactorConfig struct {
+	// The multi-factor config state.
+	State MultiFactorConfigState `json:"state,omitempty"`
 	// A slice of pointers to ProviderConfig structs, each outlining the specific second factor authorization method.
 	ProviderConfigs []*ProviderConfig `json:"providerConfigs,omitempty"`
 }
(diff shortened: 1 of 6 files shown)


pragatimodi:
Hi @ynaka81 

Thanks for contributing to Firebase SDKs. 
May I know the reason for raising this PR? Is it blocking any functionality on your end? 
We did not add it to Golang SDKs because `MultiFactorConfig.State` is only for enabling providers listed in `EnabledProviders`. It should not affect any providers in the `ProviderConfigs` list as they have an independent `State` field.

ynaka81:
Thank you for your response @pragatimodi 

Yes, I want to enable/disable "Multi-Factor Authentication" (the following image) with the back-end service written in Go. Actually, I tried to enable/disable this setting by `State` field in the `ProviderConfigs` list but I could not.

Screenshot from 2023-07-25 11-17-56

In Node SDK, it looks like it is supported by `multiFactorConfig: { state: 'ENABLED' }` according to this document https://cloud.google.com/identity-platform/docs/multi-tenancy-managing-tenants#creating_a_tenant. That's why I thought it would be good to add the same interface to the Go SDK. I also asked your support team whether there is any way to enable/disable this setting and I was confident that you are planning to support all SDKs from their response.

> Unfortunately, currently managing the Multi-factor authentication can only be achieved by using Node.js Admin SDK. You can find a full table of supported features by each SDK in the documentation [here](https://cloud.google.com/identity-platform/docs/multi-tenancy-managing-tenants#supported_features). As you can see, no SDK other than Node can currently manage multi-factor features. Recently, there have been several changes and updates to many features of Firebase Authentication and I think that our engineers wanted to have at least one SDK that supports all of them for now.
> I realize that this is not the answer you were hoping for. However, as an action for your case, I’ll escalate your suggestion to our engineers. I believe that the MFA support will be added to all of the SDKs in the future, but hopefully a Feature Request from one of our users will expedite this process. :)

If there is a workaround to enable/disable this setting in the current SDK or if I missed something, please let me know. I will follow those.

PR #573, opened 2023-07-24
feat(auth): Add `TotpInfo` field to `UserRecord`

pragatimodi:
Original [PR](https://github.com/firebase/firebase-admin-go/pull/558) based off `master`
This is based off `dev` branch.

### DIFF
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -91,7 +91,7 @@ type MultiFactorInfo struct {
 	DisplayName          string
 	EnrollmentTimestamp  int64
 	FactorID             string
-	PhoneNumber          string `Deprecated:"Use PhoneMultiFactorInfo instead"`
+	PhoneNumber          string // Deprecated: Use PhoneMultiFactorInfo instead
 	PhoneMultiFactorInfo *PhoneMultiFactorInfo
 	TOTPMultiFactorInfo  *TOTPMultiFactorInfo
 }
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -70,15 +70,15 @@ type multiFactorInfoResponse struct {
 	EnrolledAt      string    `json:"enrolledAt,omitempty"`
 }
 
-// TOTPInfo describes a server side user enrolled second totp factor.
+// TOTPInfo describes a user enrolled second TOTP factor.
 type TOTPInfo struct{}
 
-// PhoneMultiFactorInfo describes a user enrolled second phone factor.
+// PhoneMultiFactorInfo describes a user enrolled in SMS second factor.
 type PhoneMultiFactorInfo struct {
 	PhoneNumber string
 }
 
-// TOTPMultiFactorInfo describes a user enrolled second totp factor.
+// TOTPMultiFactorInfo describes a user enrolled in TOTP second factor.
 type TOTPMultiFactorInfo struct{}
 
 type multiFactorEnrollments struct {
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -713,7 +713,6 @@ func validateAndFormatMfaSettings(mfaSettings MultiFactorSettings, methodType st
 				multiFactorInfo.PhoneMultiFactorInfo = &PhoneMultiFactorInfo{
 					PhoneNumber: multiFactorInfo.PhoneNumber,
 				}
-				fmt.Println("`PhoneNumber` is deprecated, use `PhoneMultiFactorInfo` instead")
 			} else {
 				// Both PhoneMultiFactorInfo and deprecated PhoneNumber are missing.
 				return nil, fmt.Errorf("\"PhoneMultiFactorInfo\" must be defined")
@@ -1121,6 +1120,7 @@ func (r *userQueryResponse) makeExportedUserRecord() (*ExportedUserRecord, error
 				DisplayName:         factor.DisplayName,
 				EnrollmentTimestamp: enrollmentTimestamp,
 				FactorID:            phoneMultiFactorID,
+				PhoneNumber:         factor.PhoneInfo,
 				PhoneMultiFactorInfo: &PhoneMultiFactorInfo{
 					PhoneNumber: factor.PhoneInfo,
 				},
--- a/auth/user_mgt.go
+++ b/auth/user_mgt.go
@@ -42,6 +42,7 @@ const (
 	createUserMethod           = "createUser"
 	updateUserMethod           = "updateUser"
 	phoneMultiFactorID         = "phone"
+	totpMultiFactorID          = "totp"
 )
 
 // 'REDACTED', encoded as a base64 string.
@@ -62,20 +63,37 @@ type UserInfo struct {
 
 // multiFactorInfoResponse describes the `mfaInfo` of the user record API response
 type multiFactorInfoResponse struct {
-	MFAEnrollmentID string `json:"mfaEnrollmentId,omitempty"`
-	DisplayName     string `json:"displayName,omitempty"`
-	PhoneInfo       string `json:"phoneInfo,omitempty"`
-	EnrolledAt      string `json:"enrolledAt,omitempty"`
+	MFAEnrollmentID string    `json:"mfaEnrollmentId,omitempty"`
+	DisplayName     string    `json:"displayName,omitempty"`
+	PhoneInfo       string    `json:"phoneInfo,omitempty"`
+	TOTPInfo        *TOTPInfo `json:"totpInfo,omitempty"`
+	EnrolledAt      string    `json:"enrolledAt,omitempty"`
+}
+
+// TOTPInfo describes a server side user enrolled second totp factor.
+type TOTPInfo struct{}
+
+// PhoneMultiFactorInfo describes a user enrolled second phone factor.
+type PhoneMultiFactorInfo struct {
+	PhoneNumber string
+}
+
+// TOTPMultiFactorInfo describes a user enrolled second totp factor.
+type TOTPMultiFactorInfo struct{}
+
+type multiFactorEnrollments struct {
+	Enrollments []*multiFactorInfoResponse `json:"enrollments"`
 }
 
 // MultiFactorInfo describes a user enrolled second phone factor.
-// TODO : convert PhoneNumber to PhoneMultiFactorInfo struct
 type MultiFactorInfo struct {
-	UID                 string
-	DisplayName         string
-	EnrollmentTimestamp int64
-	FactorID            string
-	PhoneNumber         string
+	UID                  string
+	DisplayName          string
+	EnrollmentTimestamp  int64
+	FactorID             string
+	PhoneNumber          string `Deprecated:"Use PhoneMultiFactorInfo instead"`
+	PhoneMultiFactorInfo *PhoneMultiFactorInfo
+	TOTPMultiFactorInfo  *TOTPMultiFactorInfo
 }
 
 // MultiFactorSettings describes the multi-factor related user settings.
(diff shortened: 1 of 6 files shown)


auth/user_mgt.go
```diff
@@ -62,20 +63,37 @@ type UserInfo struct {
...
-	UID                 string
-	DisplayName         string
-	EnrollmentTimestamp int64
-	FactorID            string
-	PhoneNumber         string
+	UID                  string
+	DisplayName          string
+	EnrollmentTimestamp  int64
+	FactorID             string
+	PhoneNumber          string `Deprecated:"Use PhoneMultiFactorInfo instead"`
```
lahirumaramba, line 94: I think the deprecated should be a comment: `// Deprecated: Use PhoneMultiFactorInfo instead`

See: https://cs.opensource.google/go/go/+/master:src/text/template/parse/node.go;l=192?q=Deprecated:%20language:go&ss=go%2Fgo

integration/auth/user_mgt_test.go
```diff
@@ -447,10 +449,12 @@ func TestCreateUserMFA(t *testing.T) {
 	defer deleteUser(user.UID)
 	var factor []*auth.MultiFactorInfo = []*auth.MultiFactorInfo{
 		{
-			UID:                 user.MultiFactor.EnrolledFactors[0].UID,
-			DisplayName:         "Spouse's phone number",
-			FactorID:            "phone",
-			PhoneNumber:         "+11234567890",
```
lahirumaramba, line 453: Since the `PhoneNumber` is deprecated and not dropped, let's keep `PhoneNumber` in tests for now to check backward compatibility

integration/auth/user_mgt_test.go
```diff
@@ -434,7 +434,9 @@ func TestCreateUserMFA(t *testing.T) {
 	tc.MFASettings(auth.MultiFactorSettings{
 		EnrolledFactors: []*auth.MultiFactorInfo{
 			{
-				PhoneNumber: "+11234567890",
```
lahirumaramba, line 437: Same as below. We should still keep this test for backward compatibility and add a new test for `PhoneMultiFactorInfo`

auth/user_mgt.go
```diff
@@ -166,18 +184,25 @@ func (u *UserToCreate) set(key string, value interface{}) *UserToCreate {
...
 func convertMultiFactorInfoToServerFormat(mfaInfo MultiFactorInfo) (multiFactorInfoResponse, error) {
-	var authFactorInfo multiFactorInfoResponse
+	authFactorInfo := multiFactorInfoResponse{DisplayName: mfaInfo.DisplayName}
 	if mfaInfo.EnrollmentTimestamp != 0 {
 		authFactorInfo.EnrolledAt = time.Unix(mfaInfo.EnrollmentTimestamp, 0).Format("2006-01-02T15:04:05Z07:00Z")
 	}
-	if mfaInfo.FactorID == phoneMultiFactorID {
-		authFactorInfo.PhoneInfo = mfaInfo.PhoneNumber
-		authFactorInfo.DisplayName = mfaInfo.DisplayName
+	if mfaInfo.UID != "" {
```
Xiaoshouzi-gh, line 191: do we need to throw exceptions when UID is empty?

auth/user_mgt.go
```diff
@@ -665,18 +691,32 @@ func validateAndFormatMfaSettings(mfaSettings MultiFactorSettings, methodType st
...
+					return nil, fmt.Errorf("the second factor \"phoneNumber\" for \"%s\" must be a non-empty E.164 standard compliant identifier string", multiFactorInfo.PhoneMultiFactorInfo.PhoneNumber)
+				}
+				// No need for the else here since we are returning from the function
+			} else if multiFactorInfo.PhoneNumber != "" {
+				// PhoneMultiFactorInfo is nil, check the deprecated PhoneNumber field
+				if err := validatePhone(multiFactorInfo.PhoneNumber); err != nil {
+					return nil, fmt.Errorf("the second factor \"phoneNumber\" for \"%s\" must be a non-empty E.164 standard compliant identifier string", multiFactorInfo.PhoneNumber)
+				}
+				// The PhoneNumber field is deprecated, set it in PhoneMultiFactorInfo and inform about the deprecation.
+				multiFactorInfo.PhoneMultiFactorInfo = &PhoneMultiFactorInfo{
```
Xiaoshouzi-gh, line 713: I am not familiar with goLang, hence asking, if PhoneMultiFactorInfo is not defined, would it throw exceptions when calling multiFactorInfo.PhoneMultiFactorInfo?

auth/user_mgt.go
```diff
@@ -665,18 +691,32 @@ func validateAndFormatMfaSettings(mfaSettings MultiFactorSettings, methodType st
...
+					PhoneNumber: multiFactorInfo.PhoneNumber,
+				}
+				fmt.Println("`PhoneNumber` is deprecated, use `PhoneMultiFactorInfo` instead")
+			} else {
+				// Both PhoneMultiFactorInfo and deprecated PhoneNumber are missing.
+				return nil, fmt.Errorf("\"PhoneMultiFactorInfo\" must be defined")
```
Xiaoshouzi-gh, line 719: should this be either PhoneMultiFactorInfo or phoneNumber has be to defined?  In the else-if block, there is a possibility PhoneMultiFactorInfo is not defined correct?

Xiaoshouzi-gh (approved):
Overall LGTM.

Optional - To gradually improve the coding practice, we can start clean up while we touching the existing code.  For example, we can start moving repeated string in test cases in constants.

auth/user_mgt.go
pragatimodi: No, as of the now `multiFactorInfo.PhoneMultiFactorInfo` is `nil` so it is getting assigned to an object of type `PhoneMultiFactorInfo` here.

auth/user_mgt.go
pragatimodi: No, since UID isn't used for `create` methods. The `validateAndFormatMfaSettings` function takes care of this.

auth/user_mgt.go
pragatimodi: Yes, but the objective is to encourage `PhoneMultiFactorInfo` to be defined since `PhoneNumber` is deprecated. I'm not sure if we should encourage `PhoneNumber` here. Open to any thoughts.

integration/auth/user_mgt_test.go
```diff
@@ -445,14 +453,25 @@ func TestCreateUserMFA(t *testing.T) {
 		t.Fatalf("CreateUser() = %v; want = nil", err)
 	}
 	defer deleteUser(user.UID)
-	var factor []*auth.MultiFactorInfo = []*auth.MultiFactorInfo{
+	var factors []*auth.MultiFactorInfo = []*auth.MultiFactorInfo{
 		{
-			UID:                 user.MultiFactor.EnrolledFactors[0].UID,
-			DisplayName:         "Spouse's phone number",
-			FactorID:            "phone",
-			PhoneNumber:         "+11234567890",
```
lahirumaramba, line 453: I think `PhoneNumber` should be in at least one test case, right?

pragatimodi: Yes, this is the expected output. The idea is, on parsing the server side output, we will always display `PhoneMultiFactorInfo` and not `PhoneNumber`. `PhoneNumber` is just for input since we want to deprecate it.

auth/user_mgt.go
```diff
@@ -333,7 +358,8 @@ func (u *UserToUpdate) validatedRequest() (map[string]interface{}, error) {
 			if err != nil {
 				return nil, err
 			}
-			req["mfaInfo"] = mfaInfo
```
lahirumaramba, line 336: Sorry for missing this, but was `mfaInfo` deprecated in the BE?

auth/user_mgt.go
```diff
@@ -665,18 +691,32 @@ func validateAndFormatMfaSettings(mfaSettings MultiFactorSettings, methodType st
...
+			} else if multiFactorInfo.PhoneNumber != "" {
+				// PhoneMultiFactorInfo is nil, check the deprecated PhoneNumber field
+				if err := validatePhone(multiFactorInfo.PhoneNumber); err != nil {
+					return nil, fmt.Errorf("the second factor \"phoneNumber\" for \"%s\" must be a non-empty E.164 standard compliant identifier string", multiFactorInfo.PhoneNumber)
+				}
+				// The PhoneNumber field is deprecated, set it in PhoneMultiFactorInfo and inform about the deprecation.
+				multiFactorInfo.PhoneMultiFactorInfo = &PhoneMultiFactorInfo{
+					PhoneNumber: multiFactorInfo.PhoneNumber,
+				}
+				fmt.Println("`PhoneNumber` is deprecated, use `PhoneMultiFactorInfo` instead")
```
lahirumaramba, line 716: Let's not use `fmt.Println` in production code. I think we can remove this as we have marked `PhoneNumber` as deprecated in the API.

auth/user_mgt.go
```diff
@@ -665,18 +691,32 @@ func validateAndFormatMfaSettings(mfaSettings MultiFactorSettings, methodType st
 				return nil, fmt.Errorf("\"uid\" is not supported when adding second factors via \"createUser()\"")
 			}
 		case updateUserMethod:
-			if multiFactorInfo.UID == "" {
```
lahirumaramba, line 668: Shouldn't we keep this check for `UID`?

auth/user_mgt.go
```diff
@@ -1075,17 +1115,27 @@ func (r *userQueryResponse) makeExportedUserRecord() (*ExportedUserRecord, error
 			enrollmentTimestamp = t.Unix() * 1000
 		}
 
-		if factor.PhoneInfo == "" {
+		if factor.PhoneInfo != "" {
+			enrolledFactors = append(enrolledFactors, &MultiFactorInfo{
+				UID:                 factor.MFAEnrollmentID,
+				DisplayName:         factor.DisplayName,
+				EnrollmentTimestamp: enrollmentTimestamp,
+				FactorID:            phoneMultiFactorID,
```
lahirumaramba, line 1123: Should we still set `PhoneNumber` here? Otherwise the results from `GetUsers()` could be a breaking change, right?

auth/user_mgt_test.go
```diff
@@ -875,9 +909,11 @@ func TestInvalidUpdateUser(t *testing.T) {
 			(&UserToUpdate{}).MFASettings(MultiFactorSettings{
 				EnrolledFactors: []*MultiFactorInfo{
 					{
-						UID:         "enrolledSecondFactor1",
-						PhoneNumber: "+11234567890",
```
lahirumaramba, line 879: Can we also add a test case for the deprecated PhoneNumber for UpdateUser, please?

auth/user_mgt_test.go
```diff
@@ -886,25 +922,16 @@ func TestInvalidUpdateUser(t *testing.T) {
...
-			(&UserToUpdate{}).MFASettings(MultiFactorSettings{
-				EnrolledFactors: []*MultiFactorInfo{
-					{
-						PhoneNumber: "+11234567890",
-						FactorID:    "phone",
-						DisplayName: "Spouse's phone number",
-					},
-				},
-			}),
-			`the second factor "uid" must be a valid non-empty string when adding second factors via "updateUser()"`,
```
lahirumaramba, line 907: Shouldn't we still keep this check for `UID`?

lahirumaramba (changes requested):
Looks great overall! Thank you! Added a few more comments.

auth/user_mgt.go
pragatimodi: Yes, I was slightly confused about whether or not it would seriously affect developers. Thanks for clarifying!

auth/user_mgt.go
pragatimodi: UID is not required while updating MFA.

auth/user_mgt.go
pragatimodi: I think it was always `mfa`. https://cloud.google.com/identity-platform/docs/reference/rest/v1/accounts/update
For ref [#530](https://github.com/firebase/firebase-admin-go/pull/530).

auth/user_mgt_test.go
pragatimodi: Responded in the other comment.

auth/user_mgt.go
lahirumaramba: Was it required before? Are we fixing a bug here?

pragatimodi: No it wasn't required, it was misunderstood and overlooked previously

auth/user_mgt.go
```diff
@@ -62,20 +63,37 @@ type UserInfo struct {
...
+	DisplayName     string    `json:"displayName,omitempty"`
+	PhoneInfo       string    `json:"phoneInfo,omitempty"`
+	TOTPInfo        *TOTPInfo `json:"totpInfo,omitempty"`
+	EnrolledAt      string    `json:"enrolledAt,omitempty"`
+}
+
+// TOTPInfo describes a server side user enrolled second totp factor.
+type TOTPInfo struct{}
+
+// PhoneMultiFactorInfo describes a user enrolled second phone factor.
```
kevinthecheung, line 76:
```suggestion
// PhoneMultiFactorInfo describes a user enrolled in SMS second factor.
```

auth/user_mgt.go
```diff
@@ -62,20 +63,37 @@ type UserInfo struct {
...
+type PhoneMultiFactorInfo struct {
+	PhoneNumber string
+}
+
+// TOTPMultiFactorInfo describes a user enrolled second totp factor.
```
kevinthecheung, line 81:
```suggestion
// TOTPMultiFactorInfo describes a user enrolled in TOTP second factor.
```

kevinthecheung (changes requested):

lahirumaramba (approved):

kevinthecheung (approved):

auth/user_mgt.go
Xiaoshouzi-gh: this check's result seems to be overwrite by line 194.

davidbudnick:
Good Afternoon Team 👋🏻 

Any status updates on getting this MR merged in?

Thanks!

Xiaoshouzi-gh (approved):

auth/user_mgt.go
pragatimodi: Thanks for catching this. Fixed it!

davidbudnick:
@pragatimodi 👋🏻 Thanks for pushing the changes to get this merged into the `dev` branch. Looking through the [CONTRIBUTING page](https://github.com/firebase/firebase-admin-go/blob/master/CONTRIBUTING.md) I did not see any details about how the release process works for this project.

Do you have any idea on when my team can see it pushed to a new version?

Thanks!

msidd:
@pragatimodi  Can you please provide ETA on the next release of SDK that has this feature. A critical release is blocked due to this feature missing in the current version of SDK

pragatimodi:
@lahirumaramba Please provide an update on the estimated release for this?

lahirumaramba:
Thanks for your patience folks! We will include this change in the planned upcoming release this week.

PR #577, opened 2023-08-12
feat(auth): add signInWithCustomToken method

dehypnosis:
Hi Team,

This patch adds a public method to call an API:
- https://cloud.google.com/identity-platform/docs/reference/rest/v1/accounts/signInWithCustomToken
- which exchanges custom token to Id token and refresh token.

Sorry if this PR is not appropriate in contribution process.

### Background
Actually, I was using this REST API by calling that with API Key via separated HTTP client as the document described.

But recently, I had to turn Firebase AppCheck for Authentication due to SMS fraud issue. With AppCheck enabled, this API stops to work with "Invalid App Check Token" error in my server.

So I forked and added this API call method to firebase admin to make it work with service account credentials. And it resolves the issue.

Please check this issue.
Thanks.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


auth/auth.go
```diff
@@ -233,6 +234,51 @@ func (c *baseClient) CustomTokenWithClaims(ctx context.Context, uid string, devC
...
+
+	resp = &SignInWithCustomTokenResponse{}
+	req := &internal.Request{
+		Method: http.MethodPost,
+		URL:    fmt.Sprintf("%s/accounts:signInWithCustomToken", c.userManagementEndpoint),
+		Body:   internal.NewJSONEntity(payload),
+	}
+	if _, err = c.httpClient.DoAndUnmarshal(ctx, req, resp); err != nil {
+		resp = nil
+		return
```
nilskch, line 277: I think the return statement here is not necessary.

nilskch (commented):
I am not a contributor, but wanted to mention that the return statement is not necessary.

vksssd (approved):

Issue #586, opened 2023-09-27
Remove the actual email value from the error message in the validateEmail function or return a handleable error instead


pepnova-9:
### [REQUIRED] Step 2: Describe your environment

  * Operating System version: Mac OS Ventura Version 13.0.1
  * Firebase SDK version: v4.12.0
  * Library version:  v4.12.0
  * Firebase Product: auth (auth, database, storage, etc)

### [REQUIRED] Step 3: Describe the problem

Currently, the `validateEmail()` function returns an error with an error message that includes the actual email value passed as the argument. The error generated is a simple `wrapError`, making it impossible to handle this error using `errors.As()` or `errors.Is()` to distinguish it from other potential errors.

In my country, Japan, email addresses are considered personal information, and we are prohibited from logging them without user confirmation. Additionally, for security reasons, we do not want to log any personal information. I believe that similar restrictions apply in other countries as well.

```
func validateEmail(email string) error {
	if email == "" {
		return fmt.Errorf("email must be a non-empty string")
	}
	if parts := strings.Split(email, "@"); len(parts) != 2 || parts[0] == "" || parts[1] == "" {
		return fmt.Errorf("malformed email string: %q", email)  // actual email value is included to error message.
	}
	return nil
}
```

Suggestions:

(1) Simply remove the actual email value from the error message:

```
func validateEmail(email string) error {
	if email == "" {
		return fmt.Errorf("email must be a non-empty string")
	}
	if parts := strings.Split(email, "@"); len(parts) != 2 || parts[0] == "" || parts[1] == "" {
		return fmt.Errorf("malformed email string) 
	}
	return nil
}
```

For instance, the `validatePassword()` function does not include the actual password value in its error message. 
```
func validatePassword(val string) error {
	if len(val) < 6 {
		return fmt.Errorf("password must be a string at least 6 characters long")
	}
	return nil
}
```

This is because passwords are considered personal and highly sensitive information. Following the same logic, we can remove the actual email value from the error message.

(2) Define a specific error and make it handleable with `errors.Is()` or `errors.As()`:

```
var ErrMalformedEmailFormat = errors.New("malformed email string")

func validateEmail(email string) error {
	if email == "" {
		return fmt.Errorf("email must be a non-empty string")
	}
	if parts := strings.Split(email, "@"); len(parts) != 2 || parts[0] == "" || parts[1] == "" {
		return ErrMalformedEmailFormat
	}
	return nil
}
```

[...]

PR #587, opened 2023-09-27
Remove the actual email value from the error message returned by validateEmail()

pepnova-9:
It fixes https://github.com/firebase/firebase-admin-go/issues/586.

This is one of my suggestions that I have provided in the above issue. I believe that no one parses and utilizes the error message returned by the validateEmail() function. If there are any considerations I need to be aware of, please let me know.

RELEASE NOTE: Remove the actual email value passed as an argument from the error message returned by validateEmail().


pepnova-9:
CLA signed

PR #592, opened 2023-11-10
chore: Update integration test project setup instructions.

jonathanedey:
Updated the integration test project setup instructions to be more accurate to the current setup flow.


egilmorez (approved):
Thanks!!

PR #601, opened 2024-01-29
[chore] Update Github action workflows to fix node version and set-output deprecation warnings

jonathanedey:
- `actions/setup-go` v3 -> v5
  - Enables caching
- `actions/checkout` v2 -> v4
- `actions/github-script` 0.9.0 -> v7
- Replaced `set-output` with `GITHUB_OUTPUT`
- Swapped order of `checkout` and `setup-go` steps to allow for cache to be loaded from `go.sum`: https://github.com/actions/setup-go?tab=readme-ov-file#caching-dependency-files-and-build-outputs


lahirumaramba (approved):
LGTM! Thanks!

PR #606, opened 2024-03-18
Update README.md

lahirumaramba:
`go install` is now used to install executables (https://go.dev/doc/go-get-install-deprecation)

Updating the README to use `go get` instead


jonathanedey (approved):
LGTM!

PR #607, opened 2024-03-19
chore: Bump Supported Go versions to 1.20, 1.21, 1.22

lahirumaramba:
- Bump Go versions to 1.20, 1.21, 1.22


jonathanedey (commented):
Do we need to update `go.mod` and `go.sum` as well?

lahirumaramba:
> Do we need to update `go.mod` and `go.sum` as well?

Not for this PR. We will do a follow up PR to upgrade dependencies, which will update go.sum and go.mod for dependencies. I updated the supported version in `go.mod`. Thanks!

jonathanedey (approved):
Sounds good! LGTM

PR #608, opened 2024-03-22
chore: update docs

matinzd:
Update README for installation


lahirumaramba:
Thanks @matinzd!
This was already fixed in https://github.com/firebase/firebase-admin-go/pull/606 and will show up after the next release.

PR #613, opened 2024-04-10
[chore] Release 4.14.0 Take 2

jonathanedey:
- `github-script` v7 needs REST APIs to be called with `github.rest` instead of `github`.

https://github.com/firebase/firebase-js-sdk/pull/6247


lahirumaramba (approved):
Thank you!

Issue #619, opened 2024-05-16
FR: add sign_in_second_factor claim to FirebaseInfo


cainelli:
### Environment

  * Firebase SDK version: v4.13.0
  * Firebase Product: auth

### Describe the problem

I would want to have access to `sign_in_second_factor` to perform extra authorization based on this claim.  It would be much cleaner if given by the SDK already instead of the workaround I'm using:

```go
type FirebaseInfo struct {
	auth.FirebaseInfo
	SignInSecondFactor string `json:"sign_in_second_factor,omitempty"`
}

func verifyMFA(claims map[string]any) error {
	b, err := json.Marshal(claims["firebase"])
	if err != nil {
		return fmt.Errorf("could not marshal firebase info: %w", err)
	}
	var firebaseInfo FirebaseInfo
	if err := json.Unmarshal(b, &firebaseInfo); err != nil {
		return fmt.Errorf("could not unmarshal firebase info: %w", err)
	}
	if firebaseInfo.SignInSecondFactor == "" {
		return fmt.Errorf("2FA enforced but not enrolled")
	}
        return nil
}
```

PR #620, opened 2024-05-16
add sign_in_second_factor claim to FirebaseInfo

cainelli:
Fixes https://github.com/firebase/firebase-admin-go/issues/619. These fields are [available](https://github.com/firebase/firebase-admin-node/blob/a37eb6cb5ad92d911d0b868f5beb8d11761bcc0a/src/auth/token-verifier.ts#L107-L118) in the node SDK.


cainelli:
I see @Xiaoshouzi-gh was assigned, could you PTAL?

cainelli:
@Xiaoshouzi-gh friendly ping

Issue #621, opened 2024-05-17
VerifyIDTokenandCheckRevoked returning error: `Could not find expiry time from HTTP headers`


timelery:
Environment
Firebase SDK version: v4.14.0
Firebase Product: auth

Describe the problem
I am receiving an error that states `Could not find expiry time from HTTP headers` when executing `VerifyIDTokenandCheckRevoked` function.

This error first started occuring this morning during token validation. Looking at the source code in the module this error seems to originate in the `token-verifier` go file inside the findMaxAge method. I have verified the token I am passing to this function is valid.

othonrm:
Same happening here.

mstanleyjr:
Seeing the same for `VerifyIdToken`

timelery:
I also have new client errors coming in during phone auth against the IOS Firebase verifyPhoneNumber method. Error: "INVALID_APP_CREDENTIAL". These errors started occuring this morning as well.

jschaf:
We're still figuring out the details, but we had a multi-hour outage with this error message. We're not quite sure how things broke.

The band-aid fix was https://github.com/arryved/firebase-admin-go/pull/1 and using the `replace` directive in `go.mod` to point to our fork. That code hasn't changed in five years, so I'm not sure what suddenly caused the breakage.

timelery:
Yeah I dont believe the issue resides in the go code but in the backend Firebase API servers. The http call that go is making for token validation is no longer accepting the header payload go is sending. I tried multiple versions with the same outcome.

armando1793:
We noticed that this has been happening only in our Cloud Run instances hosted in the us-west-2 region

We first noticed symptoms of this issue around March 17, 4am GMT+8 when we were trying to refactor some of our usage of the firebase auth go sdk. We attributed the symptoms to dev error because when we would route traffic to our staging cloud run instances to an older version the issue would disappear.

We noticed the issue again on March 17, 11am GMT+8 happening to an unrelated feature from the one I mentioned above. By around March 17, 3pm GMT+8 we noticed that the issue was happening on our cloud run instances in production despite no new revisions being deployed for the last several days. This was when we started looking into the problem as a firebase issue. 

As of March 17, between 6-7pm GMT+8, we were able to use the SDK to validate client tokens via our local machines in the Philippines. But when we would try the same function inside our cloud run instances, the code would fail. Client tokens would not be validated inside the cloud run instance but would be perfectly ok on our local machines

Our findings:

[...]

PR #623, opened 2024-05-22
fix: fail gracefully when no cache headers are available

lahirumaramba:
Resolves: #621

Avoid throwing errors when parsing cache control headers for public keys and set the cache duration to 0 (equivalent to no cache) when the cache headers cannot be parsed.


auth/token_verifier.go
```diff
@@ -476,19 +473,20 @@ func parsePublicKey(kid string, key []byte) (*publicKey, error) {
 	return &publicKey{kid, pk}, nil
 }
 
-func findMaxAge(resp *http.Response) (*time.Duration, error) {
+func findMaxAge(resp *http.Response) *time.Duration {
+	duration := time.Duration(0) * time.Second
```
DellaBitta, line 477: Can this line go with the default return value at the bottom of the function?

hsubox76 (approved):

auth/token_verifier.go
lahirumaramba: Thanks! Moved the default value to the bottom

jonathanedey (approved):
LGTM

DellaBitta (approved):

Xiaoshouzi-gh (approved):
LGTM.  Thank you so much for the quick response Lahiru!

qq - r we able to test this against a real response?

lahirumaramba:
> LGTM. Thank you so much for the quick response Lahiru!
> 
> qq - r we able to test this against a real response?

Yeah, the integrations tests are all green!

PR #631, opened 2024-06-11
feat(app): allow configuring firestore instance id

aexvir:
👋🏻 hey there
I'm trying to use the firebase sdk for my app but I'm not using the default `(default)` firestore db instance, which seems to be the only supported option by this sdk

this mr introduces a new option, `firestoreID` configurable like the others, via json or via explicit values in the `Config` struct, and passes this value to the `NewClientWithDatabase` firestore constructor

if no value is supplied, the default db is still targeted, so this shouldn't incur any breaking change to existing users, moreover, there's no hardcoded values, I'm using the public variable exposed by the firestore package

I've also updated the tests to not only verify that the constructor succeeds, but also that the correct database is being used when e.g. building document refs


vksssd (approved):

vksssd (approved):

vksssd (approved):

vksssd (approved):

aexvir:
@lahirumaramba sorry for the ping, do you think someone could review this anytime soon?

PR #644, opened 2024-10-07
Support sign_in_attributes for SAML

SpringMT:
related issue : https://github.com/firebase/firebase-admin-go/issues/650

The document ([Signing in users with SAML](https://cloud.google.com/identity-platform/docs/web/saml) )  states the following:

> To retrieve the user attributes associated with the SAML provider, use the firebase.sign_in_attributes claim in the ID token. Make sure to verify the ID token using the Admin SDK when you send it to your server.

So, in this PR, `firebase-admin-go` supports `sign_in_attributes`.


SpringMT:
Should I create an issue for this PR first?

PR #638, opened 2024-08-06
feat(fcm): enable direct_boot_ok as bool in Android config

akankshaJaiswal8:
### Discussion

  * https://github.com/firebase/firebase-admin-go/issues/634

### Testing

  go test -test.short firebase.google.com/go/...         
?       firebase.google.com/go/v4/errorutils    [no test files]
?       firebase.google.com/go/v4/integration/internal  [no test files]
?       firebase.google.com/go/v4/snippets      [no test files]
ok      firebase.google.com/go/v4       3.946s
ok      firebase.google.com/go/v4/appcheck      1.608s
ok      firebase.google.com/go/v4/auth  4.664s
ok      firebase.google.com/go/v4/auth/hash     3.035s
ok      firebase.google.com/go/v4/db    2.847s
ok      firebase.google.com/go/v4/iid   2.115s
ok      firebase.google.com/go/v4/integration/auth      4.185s
ok      firebase.google.com/go/v4/integration/db        4.624s
ok      firebase.google.com/go/v4/integration/firestore 5.664s
ok      firebase.google.com/go/v4/integration/iid       5.171s
ok      firebase.google.com/go/v4/integration/messaging 6.141s
ok      firebase.google.com/go/v4/integration/storage   6.411s
ok      firebase.google.com/go/v4/internal      7.386s
ok      firebase.google.com/go/v4/messaging     6.563s
ok      firebase.google.com/go/v4/storage       7.937s

### API Changes

  * This is not an API change but a change in existing AndroidConfig as suggested in the firebase official doc: https://firebase.google.com/docs/cloud-messaging/send-message?_gl=1*1o54gcz*_up*MQ..*_ga*NTk2NDY5NTMuMTcyMDY0NjAyMg..*_ga_CW55HF8NVT*MTcyMDY0NjAyMi4xLjAuMTcyMDY0NjAyMi4wLjAuMA..#send_using_the_fcm_v1_http_api

PR #645, opened 2024-10-21
fix(fcm): Change `DirectBootOK` per Go initialisms

lahirumaramba:
* Change `DirectBootOk` to `DirectBootOK` per Go initialisms
* A followup of #638 

go/admin-fcm-directboot


jonathanedey (approved):
LGTM!

PR #648, opened 2024-10-23
chore: Add X-Goog-Api-Client header to Auth and FCM requests

lahirumaramba:
Will be included in `v4.15.0`

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"runtime"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


auth/auth.go
```diff
@@ -137,6 +138,7 @@ func NewClient(ctx context.Context, conf *internal.AuthConfig) (*Client, error)
 	hc.CreateErrFn = handleHTTPError
 	hc.Opts = []internal.HTTPOption{
 		internal.WithHeader("X-Client-Version", fmt.Sprintf("Go/Admin/%s", conf.Version)),
+		internal.WithHeader("x-goog-api-client", fmt.Sprintf("gl-go/%s fire-admin/%s", runtime.Version(), conf.Version)),
```
jonathanedey, line 141: The output of `runtime.Version()` has a prefix. Example: `go1.21` https://pkg.go.dev/runtime#Version

Would this affect parsing in our regex?

jonathanedey (approved):
LGTM with one comment on getting the go version.

auth/auth.go
lahirumaramba: Good catch! I think that should be okay as it will be consistent everywhere. Let's confirm this offline with Vinay. Thanks!

lahirumaramba: Removed the prefix. Thanks!

PR #680, opened 2025-03-26
SSRC Client Setup and Fetch Implementation 

pijushcs:

### Discussion

  * Read the contribution guidelines (CONTRIBUTING.md).
  * If this has been discussed in an issue, make sure to link to the issue here. 
    If not, go file an issue about this **before creating a pull request** to discuss.

### Testing

  * Make sure all existing tests in the repository pass after your change.
  * If you fixed a bug or added a feature, add a new test to cover your code.

### API Changes

    can discuss it together.

### DIFF
--- a/remoteconfig/remoteconfig.go
+++ b/remoteconfig/remoteconfig.go
@@ -12,8 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package for the clients to use Firebase Remote Config with Go.
-
+// Package remoteconfig for the clients to use Firebase Remote Config with Go.
 package remoteconfig
 
 import (
@@ -28,8 +27,8 @@ import (
 )
 
 const (
-	defaulBaseUrl         = "https://firebaseremoteconfig.googleapis.com"
-	firebaseClientHeader  = "X-Firebase-Client"
+	defaultBaseURL       = "https://firebaseremoteconfig.googleapis.com"
+	firebaseClientHeader = "X-Firebase-Client"
 )
 
 // Client is the interface for the Remote Config Cloud service.
@@ -37,7 +36,7 @@ type Client struct {
 	*rcClient
 }
 
-// NewClient initializes a RemoteConfigClient with app-specific detail and a returns a 
+// NewClient initializes a RemoteConfigClient with app-specific detail and a returns a
 // client to be used by the user.
 func NewClient(ctx context.Context, c *internal.RemoteConfigClientConfig) (*Client, error) {
 	if c.ProjectID == "" {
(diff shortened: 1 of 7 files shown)


remoteconfig/remoteconfig.go
```diff
@@ -0,0 +1,120 @@
...
+	version    string
+}
+
+func newRcClient(client *internal.HTTPClient, conf *internal.RemoteConfigClientConfig) *rcClient {
+	goVersion := strings.TrimPrefix(runtime.Version(), "go")
+	version := fmt.Sprintf("fire-admin-go/%s", conf.Version)
+	client.Opts = []internal.HTTPOption{
+		internal.WithHeader(firebaseClientHeader, version),
+		internal.WithHeader("X-Firebase-ETag", "true"),
+		internal.WithHeader("x-goog-api-client", fmt.Sprintf("gl-go/%s fire-admin/%s", goVersion, conf.Version)),
```
lahirumaramba, line 70: Can we update this to use `internal.GetMetricsHeader(conf.Version)` instead? please see the link for an example: https://github.com/firebase/firebase-admin-go/blob/570427a0f270b9adb061f54187a2b033548c3c9e/messaging/messaging.go#L900

remoteconfig/server_template.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return string(jsonServerTemplate), nil
+}
+
+// Evaluate processes the cached template data with a condition evaluator
+// based on the provided context.
+func (s *ServerTemplate) Evaluate(context map[string]interface{}) *ServerConfig {
+	// TODO: Write ConditionalEvaluator for evaluating
+	configMap := make(map[string]value)
```
lahirumaramba, line 131: I'd prefer us not to commit incomplete functions if we can avoid it. You can add `Evaluate` in a follow up PR with the full implementation instead of merging a placeholder here.

lahirumaramba (approved):
Thanks! Overall LGTM! added a few comments

remoteconfig/server_template.go
Namyalg: Noted, I will be adding an implementation for Evaluate in subsequent PRs

remoteconfig/remoteconfig.go
pijushcs: Yes, sure. Updated the request header to use this instead.

remoteconfig/server_template.go
pijushcs: Makes sense. Updated the PR to remove all reference to conditions and condition evaluation placeholders. We will raise two more incremental PRs on top of this covering percentage evaluation and custom signals evaluation conditions.

This PR will only for default parameter evaluation which doesn't require conditions - so all the functions are completed with respect to parameter evaluation (without conditions).

Issue #681, opened 2025-03-31
Add ActionCodeSettings.linkDomain


graeme-verticalscope:
The `ActionCodeSettings` struct is missing the [linkDomain](https://firebase.google.com/docs/reference/admin/node/firebase-admin.auth.actioncodesettings.md#actioncodesettingslinkdomain) field.

The `linkDomain` field is required to migrate off of dynamic links, it's part of the [migration guide](https://firebase.google.com/docs/auth/android/email-link-migration?#customize-mobile-links) for using a custom Firebase Hosting domain

Xiaoshouzi-gh:
Hi @graeme-verticalscope do you want to create a pull request with cherry-pick against master?  We can review it soon.

jonathanedey:
Hey folks, this is now available in [v4.18.0](https://github.com/firebase/firebase-admin-go/releases/tag/v4.18.0).

PR #682, opened 2025-04-09
Add LinkDomain to ActionCodeSettings

graeme-verticalscope:
### Discussion

Add the `LinkDomain` field to the `ActionCodeSettings` struct, this new field enables migrating off of firebase dynamic links, which is deprecated and will be unsupported soon. 

See issue https://github.com/firebase/firebase-admin-go/issues/681

### Testing

Unit test is updated to check that `linkDomain` is set correctly in the action code settings map.

I haven't tested the backend service to confirm it will accept this new parameter, but it looks like this should be the only change required, since the nodeJS SDK already includes the `linkDomain` field.

### API Changes

N/A

### DIFF
--- a/auth/email_action_links.go
+++ b/auth/email_action_links.go
@@ -31,6 +31,7 @@ type ActionCodeSettings struct {
 	AndroidPackageName    string `json:"androidPackageName,omitempty"`
 	AndroidMinimumVersion string `json:"androidMinimumVersion,omitempty"`
 	AndroidInstallApp     bool   `json:"androidInstallApp,omitempty"`
+	LinkDomain            string `json:"linkDomain,omitempty"`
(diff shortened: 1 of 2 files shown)


graeme-verticalscope:
Could someone take a look at this PR? (and related issue)

Firebase dynamic links will be shutdown on Aug 25, and I am hoping to be migrated off of dynamic links by the end of April, to ensure a comfortable buffer time before the shutdown deadline.

graeme-verticalscope:
Less then 2 months until dynamic links are shutdown. Note that this change does work :) I think it should really be in the official SDK before you shutdown dynamic links.

Xiaoshouzi-gh (approved):

graeme-verticalscope:
Closing this, since there is more recent activity on #703

PR #684, opened 2025-04-10
Implementation of Evaluate (percent condition only)

Namyalg:

- Server template types with a detailed description are added in `server_template_types.go`
- `server_template` contains the logic that evaluates the template to return the config.
- `condition_evaluator` contains the implementation of the condition type `User in Random Percentage`, custom signals will be added subsequently.


pijushcs (approved):
LGTM. You may update the description to reflect the refactoring changes as well on the final PR.

remoteconfig/server_template.go
```diff
@@ -26,21 +28,17 @@ import (
...
+	// Version information for the current Remote Config template.
+	Version version `json:"version,omitempty"`
 
-// parameter stores the representation of a template parameter.
-type parameter struct {
-	DefaultValue struct {
-		Value string `json:"value"`
-	} `json:"defaultValue"`
+	// Current Remote Config template ETag.
+	ETag string `json:"etag,omitempty"`
```
pijushcs, line 41: We can probably remove the json tag for ETag since it's not part of the actual response that is converted to JSON.

remoteconfig/server_config.go
```diff
@@ -38,39 +39,50 @@ type value struct {
...
 // NewServerConfig creates a new ServerConfig instance.
-func NewServerConfig(configValues map[string]value) *ServerConfig {
-	return &ServerConfig{ConfigValues: configValues}
+func newServerConfig(configValues map[string]value) *ServerConfig {
+	return &ServerConfig{configValues: configValues}
 }
 
 // GetBoolean returns the boolean value associated with the given key.
+//
+// It returns true if the string value is "1", "true", "t", "yes", "y", or "on" (case-insensitive).
```
pijushcs, line 61: suggestion: we can probably add the comments in this file as a separate PR that will only have refactoring / readability related changes.

remoteconfig/server_template.go
Namyalg: Have removed `omitempty`, `etag` is required while we marshal it, and expose the template via `toJSON`. (Verified that the response of `toJSON` is consistent with Node and Python)

PR #686, opened 2025-04-11
Part I - Implementation of Evaluate (percent condition)

Namyalg:

- Server template types with a detailed description are added in `server_template_types.go`
- `server_template` contains the logic that evaluates the template to return the config.
- `condition_evaluator` contains the implementation of the condition type `User in Random Percentage`, custom signals will be added subsequently.

### DIFF
--- a/remoteconfig/condition_evaluator.go
+++ b/remoteconfig/condition_evaluator.go
@@ -1,0 +1,131 @@
+// Copyright 2025 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package remoteconfig
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"log"
+	"math/big"
+)
+
+type conditionEvaluator struct {
+	evaluationContext map[string]any
+	conditions        []namedCondition
+}
+
+const (
+	maxConditionRecursionDepth = 10
+	randomizationID            = "randomizationID"
+	rootNestingLevel           = 0
+	totalMicroPercentiles      = 100_000_000
+)
+
+const (
+	lessThanOrEqual = "LESS_OR_EQUAL"
+	greaterThan     = "GREATER_THAN"
+	between         = "BETWEEN"
+)
+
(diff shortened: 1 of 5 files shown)


lahirumaramba (approved):

remoteconfig/condition_evaluator.go
```diff
@@ -0,0 +1,131 @@
...
+	evaluatedConditions := make(map[string]bool)
+	for _, condition := range ce.conditions {
+		evaluatedConditions[condition.Name] = ce.evaluateCondition(condition.Condition, rootNestingLevel)
+	}
+	return evaluatedConditions
+}
+
+func (ce *conditionEvaluator) evaluateCondition(condition *oneOfCondition, nestingLevel int) bool {
+	if nestingLevel >= maxConditionRecursionDepth {
+		log.Println("Maximum recursion depth is exceeded.")
```
lahirumaramba, line 52: Just curious: are we okay to log these instead of throwing?

remoteconfig/condition_evaluator.go
```diff
@@ -0,0 +1,131 @@
...
+	return false
+}
+
+func computeInstanceMicroPercentile(seed string, randomizationID string) uint32 {
+	seedPrefix := ""
+	if len(seed) > 0 {
+		seedPrefix = fmt.Sprintf("%s.", seed)
+	}
+	stringToHash := fmt.Sprintf("%s%s", seedPrefix, randomizationID)
+
```
lahirumaramba, line 120: How long the seed and randomizationID can be? Would it make sense to optimize this with a String builder?

```go
var sb strings.Builder
if len(seed) > 0 {
  sb.WriteString(seed) // More efficient than fmt.Sprintf for simple appends
  sb.WriteRune('.')    // Efficient for single characters
}
sb.WriteString(randomizationID)
return sb.String()
```
Just a thought, It might not make sense if the strings are not significantly long :)

Namyalg: `randomizationID` is passed by the user; could be arbitrarily long. 
Will update the PR to use the builder pattern.

remoteconfig/condition_evaluator.go
Namyalg: These log statements have been added as supporting statements to the early `false` returns; more of an implementation detail.

Additionally, wanted it to be consistent with the [Node](https://github.com/firebase/firebase-admin-node/blob/c3090575682fd09957ae675ea6024ddab2d09406/src/remote-config/condition-evaluator-internal.ts#L63) and [Python](
https://github.com/firebase/firebase-admin-python/blob/bde3fb0134b2f84d789cff47b932c07981d6565b/firebase_admin/remote_config.py#L363) SDK implementations.

However, if it is preferred that we throw rather than log; we will make the change 👍🏻

remoteconfig/condition_evaluator.go
Namyalg: Have addressed in the PR where minor refactors have been made - https://github.com/firebase/firebase-admin-go/commit/8649eaae0d8252017fe459f242ddec8c2dc74bac

PR #691, opened 2025-04-18
Part II - Implementation of Evaluate (custom signals condition)

Namyalg:

`condition_evaluator` contains the implementation of the condition type `Custom signals`. Evaluation for strings, numbers and semantic version operators, along with test cases have been added.

### DIFF
--- a/remoteconfig/condition_evaluator.go
+++ b/remoteconfig/condition_evaluator.go
@@ -16,9 +16,13 @@ package remoteconfig
 
 import (
 	"crypto/sha256"
+	"encoding/json"
 	"fmt"
 	"log"
 	"math/big"
+	"regexp"
+	"strconv"
+	"strings"
 )
 
 type conditionEvaluator struct {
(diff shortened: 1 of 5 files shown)


remoteconfig/condition_evaluator.go
```diff
@@ -121,11 +157,196 @@ func computeInstanceMicroPercentile(seed string, randomizationID string) uint32
...
+		return compareSemanticVersion(customSignalCondition.TargetCustomSignalValues[0], actualValue, func(result int) bool { return result >= 0 })
+	}
+	log.Printf("Unknown custom signal operator: %s\n", customSignalCondition.CustomSignalOperator)
+	return false
+}
+
+func (cs *customSignalCondition) isValid() error {
+	if cs.CustomSignalOperator == "" || cs.CustomSignalKey == "" || len(cs.TargetCustomSignalValues) == 0 {
+		return errInvalidCustomSignal
+	}
```
lahirumaramba, line 230: Should we validate `For numeric and semantic version operators, this will have exactly ONE target value.` here?
Something like below, if that makes sense?

```go
case numericLessThan, numericLessThanEqual, /* ... other numeric/semantic ops ... */:
    if len(cs.TargetCustomSignalValues) != 1 {
        return fmt.Errorf("operator %s requires exactly one target value, got %d", cs.CustomSignalOperator, len(cs.TargetCustomSignalValues))
    }
```

Namyalg: This validation is already currently done by the console, we can be assured that we will receive only one value (not zero, not more than one) for the case of semantic and numeric version operators.

For the case of string signals, a list of target custom signals is possible since comma-separated values are allowed
image


While for the case of numeric, not entering a value is not allowed (thus, the case where the target custom signal is empty should not occur)

image

Similarly, comma separated values (that would cause the length of the target values to be greater than 1 is not possible)

image

PS : The button `Save Condition` is disabled when there are errors

PR #695, opened 2025-05-13
fix(fcm): Optimize `SendEachInBatch` with worker pool

lahirumaramba:
The `sendEachInBatch` function, used for sending FCM messages individually within a batch, previously created a new goroutine for each message. This could lead to high CPU usage for large batches.

This commit refactors `sendEachInBatch` to use a fixed-size pool of concurrent operations (10 operations) to manage message sending. This limits the number of active goroutines, reducing CPU overhead and improving resource utilization.

Key changes:
- Implemented early validation: All messages are now validated upfront. If any message is invalid, the function returns an error immediately without attempting to send any messages.
- Introduced a mechanism for managing concurrent operations: Manages a fixed number of goroutines to process message sending tasks. Messages are distributed via channels.
- Ensured result ordering: The order of responses in `BatchResponse.Responses` correctly matches the order of the input messages, even with concurrent processing.
- Updated unit tests in `messaging_batch_test.go` to comprehensively cover the new implementation, including scenarios for varying batch sizes, partial failures, early validation, and response ordering.
- Confirmed that the existing HTTP client continues to leverage HTTP/2.

### DIFF
--- a/messaging/messaging_batch.go
+++ b/messaging/messaging_batch.go
@@ -165,51 +165,83 @@ func (c *fcmClient) sendEachInBatch(ctx context.Context, messages []*Message, dr
 		return nil, fmt.Errorf("messages must not contain more than %d elements", maxMessages)
 	}
 
-	var responses []*SendResponse = make([]*SendResponse, len(messages))
-	var wg sync.WaitGroup
-
 	for idx, m := range messages {
 		if err := validateMessage(m); err != nil {
 			return nil, fmt.Errorf("invalid message at index %d: %v", idx, err)
 		}
-		wg.Add(1)
-		go func(idx int, m *Message, dryRun bool, responses []*SendResponse) {
-			defer wg.Done()
-			var resp string
-			var err error
-			if dryRun {
-				resp, err = c.SendDryRun(ctx, m)
-			} else {
-				resp, err = c.Send(ctx, m)
-			}
-			if err == nil {
-				responses[idx] = &SendResponse{
-					Success:   true,
-					MessageID: resp,
-				}
-			} else {
-				responses[idx] = &SendResponse{
-					Success: false,
-					Error:   err,
-				}
-			}
-		}(idx, m, dryRun, responses)
 	}
-	// Wait for all SendDryRun/Send calls to finish
-	wg.Wait()
+
+	const numWorkers = 10
+	jobs := make(chan job, len(messages))
+	results := make(chan result, len(messages))
+
+	responses := make([]*SendResponse, len(messages))
+
+	for w := 0; w < numWorkers; w++ {
+		go worker(ctx, c, dryRun, jobs, results)
+	}
+
+	for idx, m := range messages {
+		jobs <- job{message: m, index: idx}
+	}
+	close(jobs)
+
+	for i := 0; i < len(messages); i++ {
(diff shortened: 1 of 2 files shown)


messaging/messaging_batch.go
```diff
@@ -165,53 +164,85 @@ func (c *fcmClient) sendEachInBatch(ctx context.Context, messages []*Message, dr
...
+		go worker(ctx, c, dryRun, jobs, results)
+	}
+
+	for idx, m := range messages {
+		jobs <- job{message: m, index: idx}
+	}
+	close(jobs)
+
+	for i := 0; i < len(messages); i++ {
+		res := <-results
```
jonathanedey, line 189: Do we need to close the request channel as well or is that automatically cleaned up.

messaging/messaging_batch.go
```diff
@@ -165,53 +164,85 @@ func (c *fcmClient) sendEachInBatch(ctx context.Context, messages []*Message, dr
...
+	responses := make([]*SendResponse, len(messages))
+
+	for w := 0; w < numWorkers; w++ {
+		go worker(ctx, c, dryRun, jobs, results)
+	}
+
+	for idx, m := range messages {
+		jobs <- job{message: m, index: idx}
+	}
+	close(jobs)
```
jonathanedey, line 186: I understand this creates 50 new workers that listen for incoming jobs until the jobs channel is closed and there are no more jobs. What happens to these workers after, do they need to be cleaned up or is that also handled by closing the jobs channel?

messaging/messaging_batch.go
```diff
@@ -165,53 +164,85 @@ func (c *fcmClient) sendEachInBatch(ctx context.Context, messages []*Message, dr
...
-					Success: false,
-					Error:   err,
-				}
-			}
-		}(idx, m, dryRun, responses)
 	}
-	// Wait for all SendDryRun/Send calls to finish
-	wg.Wait()
+
+	const numWorkers = 50
```
jonathanedey, line 173: Would raising this to 100 to match the backend max concurrent stream limit be something we consider.
I think 50 is reasonable considering this multiplies per `sendEachInBatch` call.

nit: Can we update the PR description to specify 50 workers per `sendEachInBatch` call

jonathanedey (commented):
Generally LGTM with a few comments. Thanks!

messaging/messaging_batch.go
lahirumaramba: That's a good point, I can run some tests with 100 workers to see if that affects performance

messaging/messaging_batch.go
lahirumaramba: No, it gets cleaned up automatically once it exists this loop

jonathanedey (approved):
LGTM!

messaging/messaging_batch.go
tmzane: @lahirumaramba @jonathanedey Hello! Would you consider making this value configurable?

lahirumaramba: Hey @tmzane could you please create a new issue (FR) in this repos and add some context to how this would useful to you, please? That is the best way to track the feature and make sure it won't get lost in the PR comments. Thanks!

PR #699, opened 2025-06-09
fix: set go.mod version to actual min version

anuraaga:
As we see in https://github.com/firebase/firebase-admin-go/pull/694, the actual minimum supported version is 1.23, but with `go 1.24` in the mod, dependent effectively require 1.24. The CI in this repo passes I think because `toolchain` causes Go 1.23 to just automatically download 1.24 and run it, effectively making the 1.23 CI workflow the same as 1.24.

The root cause of the go version is a transitive dependency on go-jose, https://github.com/go-jose/go-jose/issues/184. I would encourage adding to that issue to explain how problematic that is. In the meantime, while keeping to 4.0.5 (the version after the CVE) can't completely protect downstream users from tools like dependabot, at least it can protect them from `go mod tidy` which is harder to have ignore rules for.

Also, we use `1.23.0` instead of `1.23` because it corresponds to the first released version of Go 1.23, while the latter includes prereleases.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module firebase.google.com/go/v4
 
-go 1.24
-
-toolchain go1.24.1
+go 1.23.0
 
 require (
 	cloud.google.com/go/firestore v1.18.0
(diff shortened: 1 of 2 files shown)


jonathanedey (approved):
LGTM, Thank you @anuraaga for helping to resolve this!

PR #703, opened 2025-07-02
Add LinkDomain to ActionCodeSettings

graeme-verticalscope:
### Discussion

Add the `LinkDomain` field to the `ActionCodeSettings` struct, this new field enables migrating off of firebase dynamic links, which is deprecated and will be unsupported soon. 

See #681, I was asked to create a second PR targeting master branch. Original PR is #682

### DIFF
--- a/auth/email_action_links.go
+++ b/auth/email_action_links.go
@@ -31,6 +31,7 @@ type ActionCodeSettings struct {
 	AndroidPackageName    string `json:"androidPackageName,omitempty"`
 	AndroidMinimumVersion string `json:"androidMinimumVersion,omitempty"`
 	AndroidInstallApp     bool   `json:"androidInstallApp,omitempty"`
+	LinkDomain            string `json:"linkDomain,omitempty"`
 	DynamicLinkDomain     string `json:"dynamicLinkDomain,omitempty"`
 }
 
--- a/auth/email_action_links_test.go
+++ b/auth/email_action_links_test.go
@@ -35,6 +35,7 @@ var testActionLinkResponse = []byte(fmt.Sprintf(testActionLinkFormat, testAction
 var testActionCodeSettings = &ActionCodeSettings{
 	URL:                   "https://example.dynamic.link",
 	HandleCodeInApp:       true,
+	LinkDomain:            "hosted.page.link",
 	DynamicLinkDomain:     "custom.page.link",
 	IOSBundleID:           "com.example.ios",
 	AndroidPackageName:    "com.example.android",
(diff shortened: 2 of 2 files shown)


Xiaoshouzi-gh (approved):
Thanks for putting this together!

graeme-verticalscope:
I guess you'll want to close either this PR or https://github.com/firebase/firebase-admin-go/pull/682. I had originally created a PR targeting dev, but was recently asked to make one targeting master

auth/email_action_links.go
```diff
@@ -31,6 +31,7 @@ type ActionCodeSettings struct {
 	AndroidPackageName    string `json:"androidPackageName,omitempty"`
 	AndroidMinimumVersion string `json:"androidMinimumVersion,omitempty"`
 	AndroidInstallApp     bool   `json:"androidInstallApp,omitempty"`
+	LinkDomain            string `json:"linkDomain,omitempty"`
 	DynamicLinkDomain     string `json:"dynamicLinkDomain,omitempty"`
```
jonathanedey: Can we also add a deprecation note for `DynamicLinkDomain`.

graeme-verticalscope:
@jonathanedey, thanks I added the error code you mentioned! 

You may want to check that the code and message are correct here:
```
	"INVALID_HOSTING_LINK_DOMAIN": {
		code:     internal.InvalidArgument,
		message:  "the provided hosting link domain is not configured or authorized for the current project",
		authCode: invalidHostingLinkDomain,
	},
```
I used the same `code` and a similar `message` to the existing dynamic link error mapping.

auth/user_mgt.go
```diff
@@ -1447,6 +1453,11 @@ var serverError = map[string]*authError{
 		message:  "the provided dynamic link domain is not configured or authorized for the current project",
 		authCode: invalidDynamicLinkDomain,
 	},
+	"INVALID_HOSTING_LINK_DOMAIN": {
+		code:     internal.InvalidArgument,
+		message:  "the provided hosting link domain is not configured or authorized for the current project",
```
jonathanedey: Good call out, lets go with the following to match the node sdk:

> the provided hosting link domain is not configured in Firebase Hosting or is not owned by the current project

jonathanedey (approved):
LGTM with one change

auth/user_mgt.go
```diff
@@ -556,6 +557,11 @@ func IsInvalidDynamicLinkDomain(err error) bool {
 	return hasAuthErrorCode(err, invalidDynamicLinkDomain)
 }
 
+// IsInvalidHostingLinkDomain checks if the given error was due to an invalid hosting link domain.
```
egilmorez, line 560: Does Go commenting support markdown?  If so, we'd probably backtick `IsInvalidHostingLinkDomain` as a literal.

jonathanedey: I don't believe it does. We have some backticks here which we should look to fix in future: https://pkg.go.dev/firebase.google.com/go/v4#NewApp

auth/user_mgt.go
graeme-verticalscope: Great! I updated the error message.

graeme-verticalscope:
Let me know if there's anything else I need to do, otherwise I'm assuming this can be approved and merged by a repository maintainer :)

jonathanedey:
That's everything, thanks again @graeme-verticalscope for putting this together!

Issue #470, opened 2021-11-25
FR: add 'VERIFY_AND_CHANGE_EMAIL' linkType for generateEmailActionLink


tyahha:
### environment

  * Operating System version: Windows 10
  * Firebase SDK version: 4.6
  * Library version: 4.6
  * Firebase Product: auth

### Describe the problem

I want to send custom email when your change email.
And I want to verify email before change email.
Firebase Client SDK have this feature, but Firebase admin sdk have not feature to generate this type email action link.
So, As other email action link types, "EMAIL_SIGNIN", "VERIFY_EMAIL", "PASSWORD_RESET", I want to add feature to generation email action link of "VERIFY_AND_CHANGE_EMAIL" type.

This is an earlier version, but a modified version is available.
https://github.com/tyahha/firebase-admin-go/commit/2d00814367245fc692bd4fcbb1785411149a1b63

 If there are no problems, rebase to the latest version, add the test code, and create a pull request.

PR #705, opened 2025-07-10
feat(auth): Add 'VERIFY_AND_CHANGE_EMAIL' linkType for generateEmailActionLink

utamori:
## Summary
- Add support for `VERIFY_AND_CHANGE_EMAIL` linkType in Firebase Auth email action links
- Implement `VerifyAndChangeEmailLink` and `VerifyAndChangeEmailLinkWithSettings` methods
- Add comprehensive tests for the new functionality including tenant integration tests

## Changes
- Added `verifyAndChangeEmail` linkType constant
- Updated `generateEmailActionLink` to accept optional `newEmail` parameter
- Added new public methods for verify and change email functionality
- Updated existing methods to use the new signature
- Added unit tests and integration tests

## Test plan
- [x] Unit tests for new functionality
- [x] Integration tests for tenant-aware scenarios
- [x] Existing tests continue to pass

This addresses the feature request in #470 by providing the ability to generate email action links for email verification and change flows.


auth/email_action_links.go
```diff
@@ -118,6 +132,14 @@ func (c *baseClient) generateEmailActionLink(
 		"email":         email,
 		"returnOobLink": true,
 	}
+
+	if linkType == verifyAndChangeEmail {
+		if newEmail == nil {
```
macastelaz: Do we also need to check for `|| *newEmail == ""` in case an empty string is passed in here?

auth/email_action_links.go
```diff
@@ -118,6 +132,14 @@ func (c *baseClient) generateEmailActionLink(
...
+			return "", errors.New("newEmail must not be nil when linkType is verifyAndChangeEmail")
```
macastelaz: nit: "newEmail must be not be empty when linkType is verifyAndChangeEmail"

auth/email_action_links_test.go
macastelaz, line 246: Should we add a test case for client.VerifyAndChangeEmailLink being called without an email for the sake of consistency?

macastelaz, line 231: Should we have a test for when VerifyAndChangeEmailLink  is called with empty/nil newEmail?

auth/email_action_links.go
```diff
@@ -92,18 +93,31 @@ func (c *baseClient) PasswordResetLink(ctx context.Context, email string) (strin
...
 // EmailSignInLink generates the out-of-band email action link for email link sign-in flows, using the action
 // code settings provided.
 func (c *baseClient) EmailSignInLink(
 	ctx context.Context, email string, settings *ActionCodeSettings) (string, error) {
-	return c.generateEmailActionLink(ctx, emailLinkSignIn, email, settings)
+	return c.generateEmailActionLink(ctx, emailLinkSignIn, email, settings, nil)
+}
+
+// VerifyAndChangeEmailLink generates the out-of-band email action link for email verification and change flows for the
+// specified email address.
```
macastelaz: nit: "// specified current email address and new email address" to clarify there are two email addresses needed for this function.

auth/email_action_links.go
```diff
@@ -92,18 +93,31 @@ func (c *baseClient) PasswordResetLink(ctx context.Context, email string) (strin
...
+// VerifyAndChangeEmailLinkWithSettings generates the out-of-band email action link for email verification and change
+// flows for the specified email address, using the action code settings provided.
+func (c *baseClient) VerifyAndChangeEmailLinkWithSettings(
+	ctx context.Context, email string, newEmail string, settings *ActionCodeSettings) (string, error) {
+	return c.generateEmailActionLink(ctx, verifyAndChangeEmail, email, settings, &newEmail)
 }
 
 func (c *baseClient) generateEmailActionLink(
-	ctx context.Context, linkType linkType, email string, settings *ActionCodeSettings) (string, error) {
+	ctx context.Context, linkType linkType, email string, settings *ActionCodeSettings, newEmail *string) (string, error) {
```
macastelaz: I wonder if it may make sense to make this a variadic function now, making the last argument something like: "opts ...emailActionLinkOption,"

emailActionLinkOption would be something like "type emailActionLinkOption func(map[string]interface{})"

and then we'd define:

"
func withNewEmail(newEmail string) emailActionLinkOption {                
        return func(payload map[string]interface{}) {                         
            payload["newEmail"] = newEmail                                    
        }                                                                     
 } 
"

With this in place, existing call sites of generateEmailActionLink can remain unchanged (no need to pass a trailing nil) .

The new call site (VerifyAndChangeEmailLinkWithSettings) can look like this: 

"
func (c *baseClient) VerifyAndChangeEmailLinkWithSettings(            
            ctx context.Context, email string, newEmail string, settings      
      *ActionCodeSettings) (string, error) {                                  
            if newEmail == "" {                                               
                return "", errors.New("newEmail must not be empty")           
            }                                                                 
            return c.generateEmailActionLink(ctx, verifyAndChangeEmail, email,
      settings, withNewEmail(newEmail))                                       
        }
"

From my vantage point, the main advantage here is scalability. The tradeoff is the code becomes a bit more verbose and I don't know of any current needs to support additional similar options any time soon, so this may fall into the YAGNI. I'd be curious to hear your thoughts on this!

auth/email_action_links.go
utamori: Thanks. I'll fix it now.

auth/email_action_links.go
utamori: Thanks. I'll fix it now.

auth/email_action_links_test.go
utamori: Thanks. I'll fix it now.

utamori: Thanks. I'll fix it now.

auth/email_action_links.go
utamori: Thanks. I'll fix it now.

auth/email_action_links.go
utamori: Thanks! I tried it out and it worked perfectly, so I'll go with your suggestion.

macastelaz (approved):
Thanks for the contribution and making the suggested changes!

macastelaz:
Sorry I didn't catch this earlier - per the contribution guidelines, can you instead merge into the dev branch instead of master (see: https://github.com/firebase/firebase-admin-go?tab=contributing-ov-file)

Thanks!

macastelaz:
> Sorry I didn't catch this earlier - per the contribution guidelines, can you instead merge into the dev branch instead of master (see: https://github.com/firebase/firebase-admin-go?tab=contributing-ov-file)
> 
> Thanks!

Ooops - I forgot I could change this myself. None-the-less, if you have future contributions to make, feel free to target the dev branch directly.

Thanks again for the contribution here!

PR #727, opened 2025-11-21
feat(auth): Add `QueryUsers` API

lahirumaramba:
This change implements the `accounts:query` functionality, providing a new `QueryUsers` method that allows searching for users with filters and sorting options.

RELEASE_NOTE: Added `QueryUsers()` API to support querying user accounts with filters, sorting, and pagination.


pashanka (approved):

jonathanedey (approved):
LGTM, Thanks!

PR #733, opened 2025-12-04
feat(firestore): Added Firestore Multi Database Support

jonathanedey:
RELEASE NOTE: Added support for multiple Cloud Firestore named databases.


integration/firestore/firestore_test.go
```diff
@@ -40,23 +65,127 @@ func TestFirestore(t *testing.T) {
...
+func TestFirestoreWithDatabaseID(t *testing.T) {
+	ctx := context.Background()
+	app, err := internal.NewTestApp(ctx, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// This test requires a non-default database to exist in the project.
+	// If it doesn't exist, this test will fail.
+	client, err := app.FirestoreWithDatabaseID(ctx, "testing-database")
```
lahirumaramba: Let's make `testing-database` a const (instead of a string here) and add a note to `CONTRIBUTING.md` to help setup integration test projects in the future.

integration/firestore/firestore_test.go
```diff
@@ -40,23 +65,127 @@ func TestFirestore(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
```
lahirumaramba, line 92: Let's add a log here explaining the custom database might not exists in the integration project. Helps with debugging if you are setting up the workflow for the first time.

integration/firestore/firestore_test.go
```diff
@@ -40,23 +65,127 @@ func TestFirestore(t *testing.T) {
...
+	snap, err := doc.Get(ctx)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !reflect.DeepEqual(snap.Data(), cityData) {
+		t.Errorf("Get() = %v; want %v", snap.Data(), cityData)
+	}
+}
+
+func TestFirestoreMultiDB(t *testing.T) {
```
lahirumaramba, line 110: This feels repetitive. Doesn't `TestFirestoreWithDatabaseID` and `TestFirestore` cover this?

integration/firestore/firestore_test.go
```diff
@@ -40,23 +65,127 @@ func TestFirestore(t *testing.T) {
...
+
+	if !reflect.DeepEqual(citySnap.Data(), cityData) {
+		t.Errorf("City Get() = %v; want %v", citySnap.Data(), cityData)
+	}
+	if !reflect.DeepEqual(movieSnap.Data(), movieData) {
+		t.Errorf("Movie Get() = %v; want %v", movieSnap.Data(), movieData)
+	}
+}
+
+func TestServerTimestamp(t *testing.T) {
```
lahirumaramba: Why do we need a separate test to check the server timestamp?

lahirumaramba (changes requested):
Thank you! Added a few comments.

integration/firestore/firestore_test.go
jonathanedey: Good point, I'll make a note to add this to the other sdks with similar firestore tests.

integration/firestore/firestore_test.go
jonathanedey: I think the error we already return here should be good enough in the event the database isn't setup wdyt?
```
rpc error: code = NotFound desc = The database testing-database does not exist for project <project-id> Please visit https://console.cloud.google.com/datastore/setup?project=<project-id> to add a Cloud Datastore or Cloud Firestore database.
```

integration/firestore/firestore_test.go
jonathanedey: This test in for verifying that two databases clients can be opened and modified at once with no impact on the other.

integration/firestore/firestore_test.go
jonathanedey: Honestly the structure of these tests were taken from the python sdk which had a special case for testing this. I'll remove it from here.

integration/firestore/firestore_test.go
lahirumaramba: Sounds good!

integration/firestore/firestore_test.go
lahirumaramba: Okay. We can keep this, but as a clarification, to me this feels like we are trying to test the cloud firestore SDK here. This should be a more fitting test for the Firestore SDK and slightly beyond the Admin SDK scope. The Admin SDK should test the part where we are setting the database ID in the Firestore SDK and everything beyond that should be tested in the Firestore SDK itself (which I think they do).

CONTRIBUTING.md
```diff
@@ -148,8 +148,12 @@ Set up your Firebase project as follows:
 
 2. Enable Firestore:
    1. Go to the Firebase Console, and select **Firestore Database** from the **Build** menu.
-   2. Click on the **Create database** button. You can choose to set up Firestore either in
-   the production mode or in the test mode.
+   2. Click on the **Create database** button and create a default database. You can choose
+   to set up Firestore either in the production mode or in the test mode.
+   > **Note:** Integration tests are run against both the default database and an additional
+   database named "testing-database".
+   3. After the bedault datbase is created, click the **Add database** button to create a
```
lahirumaramba:
```suggestion
   3. After the default database is created, click the **Add database** button to create a
```

lahirumaramba (approved):
Thanks! LGTM!
We should be good once we get a TW review.

markarndt (approved):

PR #750, opened 2026-03-02
feat(fcm): Add support for bandwidth constrained and restricted satellite APIs

jonathanedey:
This change introduces two new configuration options to the `AndroidConfig` object: `BandwidthConstrainedOK` and `RestrictedSatelliteOK`. These boolean options allow developers to specify whether a message should be allowed to be delivered when the receiving device is in a bandwidth-constrained state or when it is connected via a restricted satellite network. Adding these properties provides greater granularity and control over message delivery behavior in resource-constrained environments.


lahirumaramba (approved):
LGTM! Thanks!

PR #751, opened 2026-03-03
chore: Revert release workflow changes to use merge commit

jonathanedey:
This pull request:
* Reverts the automated release workflows to a previous structure that relies on merging pull requests directly via GitHub actions and creating the release tag in a single workflow.
* Updates release actions to trigger on qualifying PR push to dev.
* Adds a missing `await` to the repository merge API call to ensure the branch is fully merged into the main branch before the release tag is generated.


.github/workflows/release.yml
```diff
@@ -61,40 +69,50 @@ jobs:
...
+    #   3. and the commit message has the title prefix '[chore] Release '.
+    if: github.event_name == 'push' &&
+      github.ref == 'refs/heads/dev' &&
+      startsWith(github.event.head_commit.message, '[chore] Release ')
 
     runs-on: ubuntu-latest
+    environment: Release
     permissions:
-      pull-requests: write
+      contents: write
```
lahirumaramba, line 83: Do we still use the default token now that we have access to a environment token?

jonathanedey: We use the default token for creating the release tag since that requires write access. Technically we could use the PAT token there but for principle of least privilege I think we should keep it as is.

lahirumaramba (approved):

.github/workflows/release.yml
lahirumaramba: Ah makes sense! Thanks!

PR #752, opened 2026-03-09
chore(snippets): updated code samples to use ADC and supported APIs

lahirumaramba:
Updated code samples in the snippets directory to use Application Default Credentials (ADC) wherever appropriate instead of requiring explicit service account keys. Also replaced the deprecated messaging APIs `SendAll` and `SendMulticast` with their modern, supported equivalents `SendEach` and `SendEachForMulticast`.

### DIFF
--- a/snippets/messaging.go
+++ b/snippets/messaging.go
@@ -129,7 +129,7 @@ func sendAll(ctx context.Context, client *messaging.Client) {
 		},
 	}
 
-	br, err := client.SendAll(context.Background(), messages)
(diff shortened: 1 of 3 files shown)


snippets/messaging.go
jonathanedey, line 109: I think we can remove these 3 snippets completely since there are no longer referenced in devsite and we already have the migrated version of the snippets.

https://firebase.google.com/docs/cloud-messaging/send/admin-sdk#go_1

Issue #476, opened 2021-12-31
Ability to override default retryconfig


prashanthjbabu:
I understand that the following PR ( https://github.com/firebase/firebase-admin-go/pull/215/files ) brings in a default retryconfig which allows upto 4 retries . I did notice a comment there where there was some discussion around overriding this default retryconfig . Although it wasnt clearly specified how to go about doing it ( via some example ) . So the question I have is :

1. Is it possible to override default retryconfig?
2. If so , how to go about doing this?

CC: @hiranya911

hiranya911:
@prashanthjbabu this is not possible today. `RetryConfig` is defined as an internal type -- https://github.com/firebase/firebase-admin-go/blob/6b40682756f19c6cf9b327843efaf06a4056a79e/internal/http_client.go#L365-L370

But, it would be useful to make it possible to configure the default retry behavior (may be via an App option or even an environment variable). @lahirumaramba FYA.

sahejsingh:
Came here to raise the same request.

I have another suggestion.

Currently the Default retry config checks for just one error (503).

But the firebase docs (https://firebase.google.com/docs/reference/fcm/rest/v1/ErrorCode) suggest this should be done for the following error codes:
- QUOTA_EXCEEDED
- UNAVAILABLE
- INTERNAL

sheregeda:
Can you provide me with any updates?

dani3lsf:
I would also be interested in updates on this topic.

PR #762, opened 2026-05-06
feat(fcm): add retry config client option support

aminpaydar:
Allow callers to pass retry policy through option.ClientOption so Messaging can honor custom retry behavior from NewApp options. When no retry option is supplied the default retry config is still applied.

Link to issue: https://github.com/firebase/firebase-admin-go/issues/476


aminpaydar:
@lahirumaramba can you see my pr? it need any changes ?

PR #766, opened 2026-06-29
chore(ci): address security findings flagged by zizmor

morganchen12:
- Applied `persist-credentials: false` to checkout steps to prevent credential leakage.
- Added explicit workflow-level `contents: read` permissions.
- Added explicit job-level permissions to `publish_release` in `release.yml`.


lahirumaramba (approved):

jonathanedey (commented):
Thinking this should be staged against dev instead.

@lahirumaramba I just noticed that the dev and master versions of the release workflows are slightly out of sync since I thought we removed the `tag_release` workflow and pr creation permissions in https://github.com/firebase/firebase-admin-go/pull/751. I think https://github.com/firebase/firebase-admin-go/commit/aa29809a9728bb91f2836efd971bdb56213f1217 may have overwritten these.

morganchen12:
Synced this PR with dev: https://github.com/firebase/firebase-admin-go/compare/dev...mc/zizmor

Does this diff look correct?

lahirumaramba:
Ah, good catch @jonathanedey. Thanks!

I changed the base to `dev`, but yeah the workflow needs fixing before we can merge this I think

lahirumaramba (changes requested):
Let's hold off on this until we figure out the workflow issues. Thanks!

PR #774, opened 2026-08-10
feat(appcheck): Verify one-time tokens for replay protection

yvonnep165:
This PR introduces the `VerifyOneTimeToken` method to the `appcheck` package, enabling stateful replay protection for Firebase App Check. 

- Implement a new method `VerifyOneTimeToken` that first performs local, stateless JWT verification followed by a stateful `POST` request to backend endpoint to consume the token.
- Add the `AlreadyConsumed` (*bool) field to the `DecodedAppCheckToken` struct.
- Plumb `option.ClientOption` from the main app initialization down to `internal.AppCheckConfig`. 
- Add unit tests.


appcheck/appcheck.go
```diff
@@ -33,6 +34,8 @@ var JWKSUrl = "https://firebaseappcheck.googleapis.com/v1beta/jwks"
 const appCheckIssuer = "https://firebaseappcheck.googleapis.com/"
 
 var (
+	verifyURLFormat = "https://firebaseappcheck.googleapis.com/v1beta/projects/%s:verifyAppCheckToken"
```
weixifan: Our `v1` endpoint is almost ready, and that would be our preferred URL when this SDK is released. Maybe we can just change this to `v1` now?

I'm also fine with leaving this alone for now and wait until our `v1` endpoint is fully published. In that case, is a TODO appropriate here?

weixifan (approved):
Thank you for the PR; LGTM with just a minor comment.

appcheck/appcheck.go
yvonnep165: Thanks for the quick review! Sure, we can switch to the `v1` endpoint as long as we can verify and run our tests against it. Currently, we only have unit tests in place so we can just change to `v1` now, but we plan to add integration tests against the live endpoint once we implement the token creation methods.

jonathanedey: One thing we should keep in mind if we do merge this as v1, is that our next release will include this whether or not the backend is ready. Lets set v1 but hold back merging until the backend is live.

jonathanedey (approved):
LGTM with one note on v1 endpoint. Thanks!

kevinthecheung (approved):

PR #785, opened 2026-09-17
feat(fcm): Migrate topic management to FCM v1 API

lahirumaramba:
Migrates topic management operations (`SubscribeToTopic` and `UnsubscribeFromTopic`) on the messaging client from the legacy Instance ID (IID) API to the FCM v1 REST API.

Key changes:
- `SubscribeToTopic` and `UnsubscribeFromTopic` now target FCM v1 endpoints (`/projects/{project_id}/registrations/{token}/topicSubscriptions`).
- Requests across multiple registration tokens are executed concurrently using a worker pool (up to 100 concurrent workers).
- Existing IID methods were renamed to `SubscribeToTopicLegacy` and `UnsubscribeFromTopicLegacy` and marked as deprecated.


OrlandriaH-G (approved):

messaging/topic_mgt.go
```diff
@@ -161,3 +163,198 @@ func handleIIDError(resp *internal.Response) error {
...
+	}
+
+	var respBody []byte
+	var status int
+	if fe, ok := err.(*internal.FirebaseError); ok && fe.Response != nil {
+		status = fe.Response.StatusCode
+		if fe.Response.Body != nil {
+			respBody, _ = io.ReadAll(fe.Response.Body)
+		}
+	}
```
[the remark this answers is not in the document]
lahirumaramba: @gemini-code-assist:
in `internal.FirebaseError`, `Response` is a standard library `*http.Response` where `StatusCode` is `int` (`Status` is a string like "404 Not Found") and `Body` is an `io.ReadCloser`. Using `.Status` or removing "io" causes compilation errors.

messaging/topic_mgt.go
```diff
@@ -161,3 +163,202 @@ func handleIIDError(resp *internal.Response) error {
...
+
+	resp, err := c.httpClient.Do(ctx, request)
+	if err == nil {
+		if resp != nil && isSubscribe && resp.Status == http.StatusConflict {
+			return true, ""
+		}
+		return true, ""
+	}
```
jonathanedey: With the above SuccessFn accepting 409s for subscribe calls do we still need to check for 409 status codes on error respones below again?

messaging/topic_mgt.go
```diff
@@ -161,3 +163,202 @@ func handleIIDError(resp *internal.Response) error {
...
+	encodedToken := url.PathEscape(token)
+	var request *internal.Request
+
+	if isSubscribe {
+		request = &internal.Request{
+			Method: http.MethodPost,
+			URL:    fmt.Sprintf("%s/projects/%s/registrations/%s/topicSubscriptions?topic_name=%s", c.fcmEndpoint, c.project, encodedToken, url.QueryEscape(topicName)),
+			Body:   internal.NewJSONEntity(map[string]interface{}{}),
+			SuccessFn: func(resp *internal.Response) bool {
+				return resp.Status == http.StatusOK || resp.Status == http.StatusConflict
```
yvonnep165: SuccessFn narrows subscribe to 200/409 only. But the SDK default is 
func HasSuccessStatus(r *Response) bool {
	return r.Status >= http.StatusOK && r.Status < http.StatusNotModified
}
in http_client.go. 
So like 201/202/204 would be reported as a failure while unsubscribe keeps the default 2xx behaviour?

messaging/topic_mgt.go
```diff
@@ -161,3 +163,202 @@ func handleIIDError(resp *internal.Response) error {
...
+			URL:    fmt.Sprintf("%s/projects/%s/registrations/%s/topicSubscriptions/%s?allow_missing=true", c.fcmEndpoint, c.project, encodedToken, url.PathEscape(topicName)),
+		}
+	}
+
+	resp, err := c.httpClient.Do(ctx, request)
+	if err == nil {
+		if resp != nil && isSubscribe && resp.Status == http.StatusConflict {
+			return true, ""
+		}
+		return true, ""
```
yvonnep165: Both branches return true, "". should we just collapses to `if err == nil { return true, "" }` then?

messaging/topic_mgt.go
```diff
@@ -161,3 +163,202 @@ func handleIIDError(resp *internal.Response) error {
...
+	if fe, ok := err.(*internal.FirebaseError); ok && fe.Response != nil {
+		status = fe.Response.StatusCode
+		if fe.Response.Body != nil {
+			respBody, _ = io.ReadAll(fe.Response.Body)
+		}
+	}
+
+	if isSubscribe && status == http.StatusConflict {
+		return true, ""
+	}
```
jonathanedey: Same as above, is this a needed check?

messaging/topic_mgt.go
```diff
@@ -161,3 +163,202 @@ func handleIIDError(resp *internal.Response) error {
...
+
+func (c *fcmClient) makeTopicManagementRequestV1(ctx context.Context, tokens []string, topic string, isSubscribe bool) (*TopicManagementResponse, error) {
+	topicName, err := validateTopicManagementArgs(tokens, topic)
+	if err != nil {
+		return nil, err
+	}
+
+	numWorkers := len(tokens)
+	if numWorkers > 100 {
+		numWorkers = 100
```
yvonnep165: Is the 100 worth extracting into a named constant? It repeats conceptually across SDKs.

yvonnep165 (approved):
LGTM with a few comments/questions!

messaging/topic_mgt.go
```diff
@@ -161,3 +163,202 @@ func handleIIDError(resp *internal.Response) error {
...
+		return false, "NOT_FOUND"
+	case http.StatusTooManyRequests:
+		return false, "RESOURCE_EXHAUSTED"
+	case http.StatusInternalServerError:
+		return false, "INTERNAL"
+	case http.StatusServiceUnavailable:
+		return false, "DEADLINE_EXCEEDED"
+	default:
+		return false, "UNKNOWN_ERROR"
+	}
```
jonathanedey: Not blocking but similar note as my other reviews. I don't think we need to dupicate code here. The call to `resp, err := c.httpClient.Do(ctx, request)` uses the default messaging client error handler, which already parses the backend response to a firebase error and should have the data we need to map the status codes to the v1 specific error codes.

jonathanedey (approved):

messaging/topic_mgt.go
lahirumaramba: Thanks! Fixed

messaging/topic_mgt.go
lahirumaramba: Thanks! Fixed

messaging/topic_mgt.go
lahirumaramba: Addressed. Thanks!

messaging/topic_mgt.go
lahirumaramba: Addressed. Thanks!

messaging/topic_mgt.go
lahirumaramba: Addressed. Thanks!

messaging/topic_mgt.go
lahirumaramba: Addressed. Thanks!

